PR #9, opened 2017-07-27
update CMS search index erry 15 minutes

reptillicus:
We might want to isolate the celerybeat container into its own, right now it is running in the workers container. If we were to spin up more workers, we would get more beat schedulers...


reptillicus:
Also, if this runs on staging environment, which uses the same elastic indexes as prod, there could be old CMS stuff put into the index.

PR #49, opened 2017-10-10
Des 72

rstijerina:
Material toasters replace old toastr templates. Navbar padding fixed.


reptillicus:
these toasts are the most

PR #65, opened 2017-11-13
added help button to workspace

Strmiska:
DES-116 : This is my first pull request! Let me know in slack if there is anything I need to do differently. (Keith Strmiska)


reptillicus:
should probably use the django `url` template tag so that it is relative to the host

xirdneh:
Agree with @reptillicus

Strmiska:
Oh! I see what you mean, I'll fix it.

PR #105, opened 2018-01-11
DES-288: Add 'My Projects' to Workspace browser

rstijerina:


reptillicus:
Just a heads up, several of the apps in the workspace will need some tweeks to work with projects. There are basically hard coded paths to the user's MyData directory in them. 

Your pal, 
Joe

xirdneh:
Btw, where are these hardcoded paths to a user's MyData directory?
Front end, Django Backend, wrapper.sh?

PR #110, opened 2018-01-18
Add functional 'nodeCount' field to job submission forms

rstijerina:
- Reintroduces `nodeCount` as an available parameter for users to edit in job submissions for parallel apps. 
- Calculates `processorsPerNode` when `nodeCount != defaultNodeCount`, overwriting `defaultProcessorsPerNode` and allowing successful submission

### DIFF
--- a/designsafe/apps/workspace/static/designsafe/apps/workspace/scripts/controllers/application-form.js
+++ b/designsafe/apps/workspace/static/designsafe/apps/workspace/scripts/controllers/application-form.js
@@ -70,7 +70,6 @@
       });
 
       $scope.resetForm = function() {
-        console.log($scope.data.app);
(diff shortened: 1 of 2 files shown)


reptillicus:
Why not make nodecount a parameter on the app? Would be less fiddling with clientside stuff?

reptillicus:
Why not make nodecount a parameter on the app? Would be less fiddling with clientside stuff?

xirdneh:
I think the idea is to go towards that. 
One of the downsides is that right now the front end is doing waaay to much.
We eventually need to move a lot of this logic to the backend.
Thoughs, @rstijerina ?

rstijerina:
Parse a nodecount parameter and figure out the cores in the wrapper script? Yeah that would help clean up client side, though this should work until we get around to updating all the apps.

PR #162, opened 2018-03-30
Task/des 461

jarosenb:
Modified the 'reindex' shell command to allow for optional random sampling with user-selected size and random seed.

example usage to populate local des-files index with 1,000 random samples from designsafe-es01.tacc.utexas.edu:9200/des-files:

`./manage.py reindex --remote-host='default' des-files des-files --sample=True --size=1000`

### DIFF
--- a/designsafe/apps/data/management/commands/reindex.py
+++ b/designsafe/apps/data/management/commands/reindex.py
@@ -22,9 +22,12 @@ class Command(BaseCommand):
                             default=True)
         parser.add_argument('--remote-host', help="Remote host where 'from_index' resides." \
                             "Use this if we need to copy data from a remote cluster." \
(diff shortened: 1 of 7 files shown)


designsafe/apps/data/management/commands/reindex.py
```diff
@@ -22,9 +22,12 @@ def add_arguments(self, parser):
                             default=True)
         parser.add_argument('--remote-host', help="Remote host where 'from_index' resides." \
                             "Use this if we need to copy data from a remote cluster." \
-                            "You can use http(s)?://<url>:<port> or a 'ES_CONNECTIONS' "\
+                            "You can use http(s)?://<url>:<port> or a 'gitES_CONNECTIONS' "\
```
reptillicus: IS that right?

reptillicus: gitES_CONNECTIONS?

PR #164, opened 2018-04-04
Add 'Data Collections' app category tab to Workspace

rstijerina:


designsafe/apps/workspace/static/designsafe/apps/workspace/scripts/controllers/application-tray.js
```diff
@@ -124,7 +124,7 @@
           //   return deferred.promise;
           // })
           .then(function(response){
-            const tabs = ['Simulation', 'Visualization', 'Data Processing', 'Utilities', 'My Apps'];
+            const tabs = ['Simulation', 'Visualization', 'Data Processing', 'Data Collections', 'Utilities', 'My Apps'];
```
reptillicus: Is there any way to consolidate the location of the categories? They are in two places at the moment, in the controller and the service. Might be able to stick them only in the service and get them from there in the controller.

rstijerina: Good point - I consolidated the tabs as a property of the service

PR #197, opened 2018-05-10
Applications fixes and app category name change to 'Partner Data Apps'

rstijerina:
And makes sure html app metadata only updates if one appId is returned


rstijerina:
@reptillicus Yeah, that would be cleaner, I'll do that.

PR #203, opened 2018-05-15
Task/des 507&463

JaffeJoseph:
Fixed lefthand nav bar highlighting when browsing into sub folders. Having issues displaying Published and Community Data when logged out though.


JaffeJoseph:
Also need to check state reload on box.com dropbox and googledrive

xirdneh:
This has been merged

PR #209, opened 2018-05-16
Task/des 520

jarosenb:
Enable search for multiple terms separated by `AND`/`OR`/`NOT`. Both the data depot and site-wide search are affected.

To get this to work, query strings are no longer wrapped in quotation marks if they contain spaces/special characters. For files, any term in the search that isn't `AND`/`OR`/`NOT` is wrapped in wildcard characters. If the user enters:

`irma OR nate OR maria`

the query string that Elasticsearch sees is: 

`*irma* OR *nate* OR *maria*`

This allows search hits on partial matches.

image
image
image
image

### DIFF
--- a/designsafe/apps/api/agave/filemanager/public_search_index.py
+++ b/designsafe/apps/api/agave/filemanager/public_search_index.py
@@ -417,15 +417,11 @@ class PublicElasticFileManager(BaseFileManager):
         projects_offset = offset
 
 
-        nees_published_query = Q('bool', must=[Q('simple_query_string', query=query_string)])
+        nees_published_query = Q('bool', must=[Q('query_string', query=query_string)])
         nees_published_search = LegacyPublicationIndexed.search()\
             .query(nees_published_query)\
             .extra(from_=offset, size=limit)
-
-        des_published_query = Q('bool', must=[
-            Q('simple_query_string', query=query_string),
-            Q({'term': {'status': status}}) 
-            ])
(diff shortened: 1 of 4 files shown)


xirdneh:
Is there a specific reason why we changed from `simple_query_string` to `query_string`?
`simple_query_string` should work the same way except that it will never raise an exception and ignores any invalid thing. For instance if somebody inputs a weird character or something like that.
I remember when we were using `query_string` we were getting some errors when people were searched for something like `NEES-2011-1088` or `rate +/-` or some unicode chard like that dreaded back tic 😠 . Things like that

Also, do we need to  wrap the query string with `*`?
I think we've been doing that forever but I don't think we actually need it.
With some of the mapping updates we should be able to get away without the `*`.
@jarosenb can you test if it works like that?
If it doesn't lets create a jira ticket so we can do some research and figure out how can we avoid using `*` on the query string.
The reason for this is because everything is slower with wildcards.

Thanks!

jarosenb:
I just played with this in local dev (querying both prod and staging ES). 

- If query strings aren't wrapped with `*` then we won't get hits at all for files unless the search query is an exact match for the filename. I messed around a bunch yesterday with the fuzziness/slop distance settings but nothing seemed to help. 

- If `simple_query_string` is used then the wildcards aren't respected at all, which is why I changed to `query_string`. Also, boolean operators don't work- searching `Maria OR Nate` gives me hits for `Maria` and `Nate` as well as a file with `or` in the name.

xirdneh:
Awesome. Thanks for the update.

PR #214, opened 2018-05-21
Task/des 533

jarosenb:
Checking the user agent on the server side using `request.META['HTTP_USER_AGENT']` has issues in production. The string retrieved from the `request` object sometimes doesn't contain the user's actual browser.

With this change, the user agent string is pulled from the browser on the javascript side with `navigator.userAgent`.


jarosenb:
Yes, the middleware is redundant now (until we figure out why we're getting bad user agent strings in prod). 

I accidentally branched this fix off of task/DES-503 (which implemented the middleware) which is why it showed up in the changes. I just did a rebase, so the diff should make more sense now.

PR #223, opened 2018-06-05
Task/des 482

jarosenb:
Added a check to make sure Stampede 2 is up before users try to submit jobs to it. This check is done using the XSEDE API: https://api.xsede.org/systems/v1/stampede2.tacc.xsede.org

Displays a warning on loading the application and disables the `Run`/`Launch` button.

screen shot 2018-06-04 at 4 57 35 pm

### DIFF
--- a/designsafe/apps/workspace/static/designsafe/apps/workspace/html/application-form.html
+++ b/designsafe/apps/workspace/static/designsafe/apps/workspace/html/application-form.html
@@ -37,6 +37,13 @@
             to another execution system such as Stampede2, or submit a ticket for assistance.
         </p>
     </div>
+    <div class="lead alert alert-warning" ng-if="data.systemDown">
+        <h2>Application Unavailable</h2>
+        <p>
+            This application's execution system ({{data.app.executionSystem}}) is not currently accepting requests. Please try again later, 
+            or submit a ticket for assistance.
+        </p>
+    </div>
     <div class="messages">
         <div ng-repeat="message in data.messages" class="alert alert-{{message.type}}">
             <h4 ng-if="message.header" ng-bind-html="message.header"></h4>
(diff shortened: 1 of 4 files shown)


designsafe/apps/workspace/static/designsafe/apps/workspace/html/application-form.html
```diff
@@ -57,6 +57,14 @@ <h2>Run <em>{{data.app.label}} <small class="version-label">ver. {{data.app.vers
           sf-options="{ supressPropertyTitles: true }"
           ng-submit="!data.submitting && onSubmit(appForm)"></form>
 
+    <div ng-if="!data.up" style="padding-top: 10px">
```
rstijerina: Look at moving this alert to be above the app form, at the same level as the data.unavailable alert.

designsafe/apps/workspace/static/designsafe/apps/workspace/scripts/services/apps-service.js
```diff
@@ -45,6 +45,30 @@
       });
     };
 
+    service.getSystemStatus = function(system) {
```
rstijerina: Service looks good. Would it make more sense to move it to the Systems service instead of the Apps service?

Also it would be good to add a docstring with an example of what was returned from these calls.

designsafe/apps/workspace/static/designsafe/apps/workspace/scripts/controllers/application-form.js
```diff
@@ -220,7 +227,8 @@
             jobData.processorsPerNode = jobData.nodeCount * ($scope.data.app.defaultProcessorsPerNode / $scope.data.app.defaultNodeCount);
           }
 
-          $scope.data.submitting = true;
+          if ($scope.data.up) {
```
rstijerina: Rather than have this here, make the form readonly if the system is down (lines 108, 125 & 132). Maybe make a readonly attribute for clarity.

PR #231, opened 2018-06-20
Task/des 194

fnets:
When a user logs in, DesignSafe now checks if the user has an Agave home directory synchronously.  If the user has one, the directory is created synchronously.  Then, the normal check_or_create task is run asynchronously.

Removed signals.py from auth, because it was no longer useful.

Removed successful account creation and login messages, because they were not appearing consistently.

### DIFF
--- a/designsafe/LoginTest.py
+++ b/designsafe/LoginTest.py
@@ -3,6 +3,7 @@ from django.test import TestCase, RequestFactory
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
 import mock
+import json
 from designsafe.apps.auth.models import AgaveOAuthToken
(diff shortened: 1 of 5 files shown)


xirdneh:
Can you make it so the `mkdir` action is only executed if the http error is a 404?
https://github.com/DesignSafe-CI/portal/pull/231/files#diff-7be4c1901cc57de93b7590a9e94b4639R169
Something like:
```python
except HTTPError as exc:
    if exc.status_code == 404:
        ag.files.manage()
```
That way the code doesn't try to create the directory if Agave is down or something like that.
Specially since a 5XX error probably takes more time to return and still the home directory will get created next time the user logs in and Agave is up.

fnets:
Made changes suggested by @xirdneh.

PR #232, opened 2018-06-20
Task/des 557: Set up indexing for projects in My Projects

jarosenb:
- Added a task to index projects when they are created or saved
- Added a periodic task to reindex all projects (works for 667/667 projects)
- NOTE: The existing `des-projects` indexes in dev/prod will need to be re-initialized for indexing to work on all projects.


xirdneh:
Can you update the `index_or_update_project` task to only take a uuid then within the task retrieve the project value and index that.
It's usually a good practice to not send the entire object as argument in tasks and instead send only a reference. This way you make sure you're not working with stale content within the task and also every parameter is serialized with JSON. Things can go wrong when serializing an entire object and those errors are really hard to find and debug.

Also, can you remove the field `teamMember` that was a mistake and some projects have it, then we realized it was a typo and we changed it. We've been paying the price of that error for a while now, hopefully we can start removing that field entirely.

Thanks! this looks great.

jarosenb:
Thanks for the feedback! I just pushed the requested changes.

PR #234, opened 2018-06-21
task/565

Strmiska:
Fixes a problem where added associated projects would not show up in project metadata consistently. Changed ui elements for adding/removing associated projects and how they appear in the metadata section.


xirdneh:
This looks great.
Although, is that `$state.reload()` necessary?
The few times I've used it it has been buggy, but maybe it's better now.

PR #239, opened 2018-06-27
DES-580: Search within projects in My Projects

jarosenb:
Implements a search within projects in My Projects (previously searches could only be run from the general projects listing).

This is done by adding a `query_string` argument to the `projects.view.data` router state and using it to display search results. An optional `params` argument was added to the `DataBrowserService.browse` method so that the query string can be passed to the backend without affecting file listings in other parts of the Data Depot.

proj_search_big


xirdneh:
Can you remove the debug statement that prints the query string?
It tends to clutter the logs a lot.
Everything else looks great, thanks!

PR #253, opened 2018-07-20
Task/des 608: Correctly index file mimetypes in Elasticsearch

jarosenb:
Adds a function to look up mimetypes using `python-magic` before indexing files. This will be necessary for any filtering based on file types. 

There are two implementations of the lookup function, depending on whether the Corral filesystem is available.


jarosenb:
Done. Directories are kind of tricky since the Agavepy `client.files.download` doesn't have a directory-specific response like `requests` does. I'm assuming `501 Server Error: Not Implemented for url` will only get thrown for directories- is this a safe assumption?

This only matters in local dev, since when `DEBUG==False` we check directly using `os.path.isdir`.

PR #257, opened 2018-08-01
Task/des 207 - Workspace: refactor backend to utilize the existing api.agave calls

letaniaferreira:
https://jira.tacc.utexas.edu/browse/DES-207

Please notice that tests.py need attention:

- All tests in the TestWorkspace class are working and run fine if the WorkspaceViewtestCase is commented out. 

- As it is right now we seem to be having 2 different types of problem supposedly related to  CMS in WorkspaceViewtestCase:

1- The mocking done to run tests in the WorkspaceViewtestCase class only allows tests to run one at a time. We need to comments out all tests in the TestWorkspace class and uncomment each test that we want to run in WorkspaceViewtestCase individually for them to work.

2- All tests in the WorkspaceViewtestCase class work if run individually, except for the 3 tests marked with the following error message: "The test below currently gives the following error message: TemplateSyntaxError: You must enable the 'sekizai.context_processors.sekizai'"


xirdneh:
Closing this one since we have to re-visit this with the work done in CEP.

PR #264, opened 2018-08-22
Task/des 590 - Send e-mail to collaborators added to projects

letaniaferreira:
https://jira.tacc.utexas.edu/browse/DES-590

send e-mail to collaborators and to copis has been fixed and added to tasks.py

In order to make the email work in dev environment we created a smtp container.


letaniaferreira:
Yeah, we can delete the smtp service, that should not be needed in prod.
No need for socket on api.tasks.py either. It can go as well.

letaniaferreira:
Changes had been made. Thank you for the comments!

PR #269, opened 2018-08-27
Task/des 149 - Generating user report

letaniaferreira:
https://jira.tacc.utexas.edu/browse/DES-149

By hitting the following endpoint : https://www.designsafe-ci.org/account/user-report/announcements, user is redirected to a page that says that report is being generated and will be sent to his home directory in a few minutes.

Report is generated using an asynchronous task and sent to user home directory when done.

Note - The option of using an endpoint instead of a button to trigger the report was considered optimal by user and Tracy is of the opinion that we should simplify the task as just one person will be using this functionality.


fnets:
This PR is completed in [here](https://github.com/DesignSafe-CI/portal/pull/281)

PR #279, opened 2018-09-19
bug/Licenses

Strmiska:
The names of the two "other" licenses were incorrectly assigned according to Maria. The titles/names should match with the correct icons and links now.


designsafe/static/scripts/ng-designsafe/html/directives/dd-listing.html
```diff
@@ -3075,14 +3075,14 @@ <h5>Other Licenses: </h5>
        <input type="checkbox" name="cc_attrib" id="license-cc_attrib" />
        <label for="license-cc_attrib">
          <img src="/static/images/licenses/cc_attrib.png" width="100px" style="margin-right:15px;">
```
rstijerina: Should just rename these files themselves to what they should be. Then you won't have to mess with the labeling.

cc_attrib.png should be cc_public.png and vice versa

Strmiska: Now that you mentioned it. I think the file names of **cc_attrib.png** and **cc_public.png** are mixed up as well. cc_public should be the zero icon and vice versa. Right now **cc_attrib.png** is the the creative commons public domain icon.

PR #281, opened 2018-09-21
Task/des 149

fnets:
This is a continuation and cleanup of the [PR for DES-149](https://github.com/DesignSafe-CI/portal/pull/269).

I have marked this with as the "Celery" branch, because it completes the task with a Celery job that writes the user_report to the user's Agave home directory.  Tim has expressed that while this is ok, he would prefer that the report is just downloaded straight to his computer.  

Since I wasn't involved with the decision, and @xirdneh (who was) is out, I would push both a branch that stores the user_report in Agave and a branch that saves the file directly to the user's local machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Improvements:
 
 - Remove `nodeCount` parameter in front end.
 - Add email notification for new users.
+- Add user report for admins.
 
 Fixes:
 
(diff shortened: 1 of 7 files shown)


fnets:
Through testing, and verifying with @xirdneh, I found that a local download of a CSV of this size will time out.  So, this is the best way to implement this task.

PR #284, opened 2018-09-26
Task/des 479

fnets:
### **This is a work in progress**

This writes `NEES` metadata to the project page in the same way that `designsafe.storage.published` metadata is written on the page. 

I've written a new state, called `publishedNeesData`, that calls the same template and controller as `publishedData`.  This was necessary because any change to the controller that would be called when selecting a NEES project would also affect the Published data listing. So, they needed to be separated. 
 
Currently, the API call `https://designsafe.dev/api/projects/publication/NEES-XXXX-XXXX.groups` is getting a 404.  I haven't been able to figure out why, because that call is made in `neesCitation` in the `data-browser-service`.


xirdneh:
This is being re-worked with all the changes to ES6.

PR #287, opened 2018-10-02
task/DES-676 uncategorized data notification and experiment sort

Strmiska:
Created a check for project categories without associated data. Previously if a user tried to proceed through the publication pipeline without uncategorized data the selection area would become unresponsive without notifying the user.

Also, the relationship tree for unpublished experimental projects should now sort alphabetically. Included custom sort capability for published projects.


Strmiska:
notification example

PR #299, opened 2018-10-11
task/DES-677 removing alerts for non-required project data

Strmiska:
- Removed some html in metadata section for Integrated Data Analysis and Integrated Reports on Simulations. ( no longer in use )
- Removed the red ! alert icons on missing fields that are not required. (definitions)
- isMissing function tweaked to additionally check Arrays ( bug related to missing Authors not displaying notification )


xirdneh:
Can you rebase/merge with master?
I think that will make coverage not decrease. If not, can you make coverage not decrease?
Thanks!

PR #307, opened 2018-10-24
DES-521 DES-541, also moved recon stuff into root static

reptillicus:
Moved recon stuff into root static, added a tooltip on the markers with the event title. 

screenshot from 2018-10-24 15-18-42


designsafe/static/scripts/ng-designsafe/services/user-service.js
```diff
@@ -22,7 +22,7 @@ export function UserService($http, $q, djangoUrl, Logging) {
      * @returns {Promise}
      */
     this.search = function (options) {
-      return $http.get(djangoUrl.reverse('designsafe_api:user_search'), {params: {q: options.q, role:options.role}})
+      return $http.get("/api/users", {params: {q: options.q, role:options.role}})
```
xirdneh: 👍

PR #309, opened 2018-10-25
DES-706: Update project breadcrumbs for Angular 1.6

jarosenb:
Project breadcrumbs work with angular 1.6+ and no longer break when the user clicks them or refreshes the page.

Breadcrumbs are updated after each project/file listing by pushing to an RxJS subject kept in the data browser service.


jarosenb:
Done.

PR #315, opened 2018-11-05
Reorganize Data Depot controllers to components

jarosenb:
Restructure controller/template pairs in the data depot into components.

This is just code reorganization without any fancy ES6 class stuff. The only real change is that some state from the projects controllers has been moved to the projects service, because scope can't be shared between components like we were doing between controllers.


designsafe/static/scripts/data-depot/components/community/community.component.js
```diff
@@ -0,0 +1,126 @@
+import _ from 'underscore';
+import CommunityTemplate from './community.component.html';
+
+export function CommunityDataCtrl($scope, $state, Django, DataBrowserService) {
```
xirdneh: Add
```
'ngInject';
```

designsafe/static/scripts/data-depot/components/my-data/my-data.component.js
```diff
@@ -1,20 +1,8 @@
 import _ from 'underscore';
+import MyDataTemplate from './my-data.component.html';
+
+export function MyDataCtrl($scope, $state, Django, DataBrowserService) {
```
xirdneh: Add `'ngInject';`

designsafe/static/scripts/data-depot/components/projects/project-root/project-root.component.js
```diff
@@ -0,0 +1,93 @@
+import _ from 'underscore';
+import projectRootTemplate from './project-root.component.html';
+
+export function ProjectRootCtrl($scope, $state, DataBrowserService, ProjectService) {
+    $scope.browser = DataBrowserService.state();
```
xirdneh: Add `'ngInject';`

designsafe/static/scripts/data-depot/components/projects/project-search/project-search.component.js
```diff
@@ -0,0 +1,95 @@
+import ProjectSearchTemplate from './project-search.component.html';
+import _ from 'underscore';
+
+function ProjectSearchCtrl($scope, $state, Django, DataBrowserService, ProjectService) {
+
```
xirdneh: Add `'ngInject';`

designsafe/static/scripts/data-depot/components/community/community.component.js
jarosenb: I've added all the `ngInject` annotations and made sure it works with `npm run build`. 

Can you clarify what you mean by "add comments on the controllers to deprecate the use of $scope" ? Those controllers still use $scope (that will hopefully change soon though), I only replaced $scope variables with service variables when state was being shared between components.

xirdneh: I mean to add an explicit comment that we need to do that. But i think by now we all know we have to do it. So, it's ok :)

PR #318, opened 2018-11-06
hotfix for recon portal, the css stuff got borked in one of those merges

reptillicus:
Still a bit of hack to get the recon CSS bundled, but it works for now...
Also removed ngAnnotate plugin from webpack config, need to update packages.json


jarosenb:
lgtm

PR #324, opened 2018-11-09
Task/des 703: Applications builder fixes

rstijerina:


designsafe/apps/applications/views.py
```diff
@@ -274,7 +274,7 @@ def call_api(request, service):
                                     'query': query
                                 }
                             })
-                            data = agave.meta.listMetadata(q=query)
+                            data = agave.meta.listMetadata(q=query, limit=0)
```
xirdneh: Why `limit=0`?

designsafe/static/scripts/applications/controllers/application-systems-role.js
```diff
@@ -1,46 +1,44 @@
 export function applicationSystemsRoleCtrl(window, angular, $, _) {
```
xirdneh: Can we re-write to be a class?

designsafe/static/scripts/applications/services/apps-service.js
```diff
@@ -1,187 +1,181 @@
...
-        method: 'GET',
-        params: {'system_id': system_id, 'path': path}
-      });
-    };
-
-    return service;
-  }]);
-
-}
+export function appsService(window, angular, $, _) {
```
xirdneh: Can we re-write this to be a class?

designsafe/static/scripts/applications/services/simple-list-service.js
```diff
@@ -1,184 +1,81 @@
-export function simpleListFactory($http, $q, $translate, djangoUrl, appIcons) {
+export function simpleListFactory($http, $q, $translate, djangoUrl, appIcons, appCategories) {
```
xirdneh: Can we re-write this to be a class?

designsafe/apps/applications/views.py
rstijerina: `limit=0` allows for unlimited responses from the query - before in the applications view we were only getting the default max

designsafe/static/scripts/applications/services/apps-service.js
rstijerina: Sure thing

designsafe/apps/applications/views.py
xirdneh: Probably better to write a method to paginate this.
Although the metadata records don't grow drastically from one day to another and we still need to re-write this entire view class.
Let me know if you have time to write a paginator for this. If not, please create a task for it so we can do it in the future, eventually.

rstijerina: I agree 100%. I'll see if I can get around to it with these class changes, but will probably have to make a task for the future - need to get to transition to core

xirdneh: 👍 Sounds good.

designsafe/static/scripts/applications/app.js
```diff
@@ -5,31 +5,32 @@ import _ from 'underscore';
...
 applicationTrayCtrl(window, angular, $, _);
 toTrusted(window, angular, $, _);
 appTranslateProvider(angular);
 appsPemsService(window, angular, $, _);
-appsService(window, angular, $, _);
 appsMultipleListService(window, angular, $, _);
 
-angular.module('designsafe').factory('SimpleList', ['$http', '$q', '$translate', 'djangoUrl', 'appIcons', simpleListFactory]);
+angular.module('designsafe')
+    .factory('SimpleList', ['$http', '$q', 'djangoUrl', 'appIcons', ($http, $q, djangoUrl, appIcons) => new SimpleList($http, $q, djangoUrl, appIcons)])
```
rstijerina: ^ Had to manually inject these dependencies for the factory

PR #325, opened 2018-11-12
Ticket body newline issue

fnets:
Added a space between ticket body and new line in DjangoRT. This solution was implemented in [CEP](https://bitbucket.org/taccaci/core-exp-portal/pull-requests/163/ticket-body-newline-issue/diff) by Carrie Arnold and ported over to DesignSafe by me.

### DIFF
--- a/designsafe/apps/djangoRT/views.py
+++ b/designsafe/apps/djangoRT/views.py
@@ -92,7 +92,7 @@ def ticketcreate(request):
             )
 
             ticket = rtModels.Ticket(subject=form.cleaned_data['subject'],
-                                     problem_description="\n  ".join(ticket_body),
+                                     problem_description="\n  ".join(ticket_body.splitlines()),
                                      requestor=form.cleaned_data['email'],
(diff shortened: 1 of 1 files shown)


designsafe/apps/djangoRT/views.py
```diff
@@ -92,7 +92,7 @@ def ticketcreate(request):
             )
 
             ticket = rtModels.Ticket(subject=form.cleaned_data['subject'],
-                                     problem_description=ticket_body,
+                                     problem_description="\n  ".join(ticket_body),
```
xirdneh: Is `ticket_body` a string or an array here?
If it's a string isn't this going to add a new line in between each char?

fnets: Yup, you're right.  I'll work on fixing it.

fnets: Added a `splitline()` call which resolves this.

fnets:
Working on it!

PR #326, opened 2018-11-13
hotfix for saving things in hazmapper

reptillicus:
Hotfix, need to revisit this soon. Am going to move hazmapper into /static soon, will take a look then.


reptillicus:
Ah wait, another bug

PR #327, opened 2018-11-13
DES-713: Add header metadata to publication view pages.

jarosenb:
This is done by modifying the header directly using `$window`.

### DIFF
--- a/designsafe/static/scripts/data-depot/index.js
+++ b/designsafe/static/scripts/data-depot/index.js
@@ -8,8 +8,9 @@ import _ from 'underscore';
 
 import './../ng-designsafe/providers';
 import './components';
+import './services';
 
-let ddModule = angular.module('ds-data', ['designsafe', 'dd.components']);
+let ddModule = angular.module('ds-data', ['designsafe', 'dd.components', 'dd.services']);
 ddModule.requires.push(
   'ui.router',
   'djng.urls', //TODO: djng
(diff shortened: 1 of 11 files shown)


designsafe/static/scripts/data-depot/index.js
```diff
@@ -480,14 +480,41 @@ function config($httpProvider, $locationProvider, $stateProvider, $urlRouterProv
...
-        'listing': ['$stateParams', 'DataBrowserService', function($stateParams, DataBrowserService){
+        'listing': ['$stateParams', '$http', 'DataBrowserService', function($stateParams, $http, DataBrowserService){
           var systemId = $stateParams.systemId || 'designsafe.storage.published';
           var filePath = $stateParams.filePath;
           DataBrowserService.apiParams.fileMgr = 'published';
           DataBrowserService.apiParams.baseUrl = '/api/public/files';
           DataBrowserService.apiParams.searchState = 'publicDataSearch';
           return DataBrowserService.browse({system: systemId, path: filePath});
         }],
+        'headerMeta': ['$http', '$window', '$filter', '$stateParams', ($http, $window, $filter, $stateParams) => {
```
xirdneh: Can we move this into the controller?
We really shouldn't do a lot on these resolve states.
I know ui-router documentation say to do something like that but then it messes up with the UX because nothing is rendered when these resolve values are being, well, resolved.
I guess we could change that and add a "Loading" message on every template but it's easier if we move all of this logic into the controller that'll take care of it.
That way we don't have to hunt down what's going on on a state's init.

jarosenb: We could, but my fear is that Google's crawler might read the `meta` tags at a point after the route resolves but before the controller populates them with content. If this isn't actually a concern I'll move this stuff into the controller.

xirdneh: That is a concern.
But we can test that once it's in dev. We would have to ask who has access to google's search console, probably @heddabedda. Or I think we can do it outside the [Google Search console.](https://search.google.com/structured-data/testing-tool/u/0/)

Sucks that apparently we can't we use [LD+JSON](https://developers.google.com/search/docs/guides/intro-structured-data).

jarosenb:
Unit tests are in. I ended up moving some code around in order to make the unit tests non-horrible:
- `PublishedComponent` is now a `$scope`-free ES6 class.
- HTTP requests and the header DOM manipulation are moved to a new `PublishedController` (with tests!)

PR #328, opened 2018-11-15
Task/des 543: Add "Connect" button to modal and action link for interactive web apps

rstijerina:


rstijerina:
This doesn't fix the job status notifications, this is just to expand webhooks for other non vnc interactive apps. I'll add a test

PR #335, opened 2018-12-04
Recon portal state aware DES-714 and DES-705

reptillicus:
Moved geo to static folder, modified the css bundling so that each
entry can have a css bundle also. Need to double/triple check that all
the necessary CSS is getting included. A lot of this is linting/indentation. 

In the process, ended up having to move the hazmapper out of the apps directory and into static (DES-705)


.eslintrc
```diff
@@ -14,6 +14,7 @@
     },
     "rules": {
         "semi"                  : [1, "always"],
+        "require-jsdoc"         : 0,
```
xirdneh: will this disable JSDoc's check?

PR #351, opened 2019-01-10
Task/DES-762 - associated prj href and table overflow

Strmiska:
Fixed associated project href binding and an issue that caused data to overflow in the published project table.


Strmiska:
Okay, should be caught up with master now.

PR #373, opened 2019-02-14
task/DES_812: componentize publication listing 

jarosenb:
- The list of publications is now a component with unit tests.
- Searching for a publication now uses the same router state as a listing, like searches in CEP. 
- Adds search for files within a publication.
- Fix breadcrumbs in the publication listing so the root doesn't try to link to a non-existent NEES pub.


xirdneh:
@jarosenb Can you take care of the conflicts here? Thanks.

PR #374, opened 2019-02-14
task/DES-684 Header created and user guide linked for Data Depot page

eliasmarkc:
data_depot.html was modified to include a header similar to the one on workspace's index.html. The new header includes a link to the Data Depot User Guide.


eliasmarkc:
screen shot 2019-02-14 at 12 48 14 pm
It's identical in style as the existing header and link on the workspace page.

PR #392, opened 2019-02-21
task/des 831

jrfreez:
Changed case in methods and functions in Geo services, models and utils to camel case.
Changed calls to those functions and methods to match.


jrfreez:
This one should be conflict free now as well: task/des 831 (#392)
Joshua

PR #394, opened 2019-02-22
Task/DES-838 - Duplicated User Fix and Edit Project Type Button

Strmiska:
Fixed an issue causing usernames to be duplicated upon project creation
Added project type selection and warning message to edit project modal


Strmiska:
I'm not sure where the creator of the project is being added to the list of teamMembers, but this is only an issue if the pi is the creator (it duplicates the name on the created project). I thought this was being handled in the views, but I had some issues when I removed the user from being added to the teamMembers section on the new project. The project metadata would be created, but even though the creator would exist as a PI or CoPI the project would not show in "My Projects"

PR #397, opened 2019-02-22
Des 570

climberfields:
I fixed the issue with the blank spot populating in the drop down for the move component. It was an issue of test code. The dropdown now has My Projects listed first for both the Move and Copy components. The code was updated in two places; data-browswer-service-move.component line 34 and Data-browser-service.js line 305 I moved ‘My Projects’ to before ‘My Data’. ‘My Data’ now at line 308.


climberfields:
screen shot 2019-03-08 at 10 48 26 am
screen shot 2019-03-08 at 10 48 06 am

PR #406, opened 2019-03-01
task/DES-370

jrfreez:
Went through design design safe directories adding .then and error to $http calls wherever it was absent in .js files. Most were in the services directories but there were others.

### DIFF
--- a/designsafe/static/scripts/applications/services/apps-service.js
+++ b/designsafe/static/scripts/applications/services/apps-service.js
@@ -12,13 +12,25 @@ export class Apps {
             method: 'GET',
             params: { q: query },
             cache: true,
-        });
+        })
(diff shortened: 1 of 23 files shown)


.gitignore
```diff
@@ -1,3 +1,6 @@
+package-lock.json
```
xirdneh: We have to track these files to make sure that all the JS files are the same everywhere.

jrfreez:
DES-370 should be resolved now (and the gitignore file as well).

Joshua

xirdneh:
@jrfreez I had to revert this PR because there's a piece missing. We need to make sure that the services that were updated are being used in the correct way.

jrfreez:
What do I need to do with it now?
Joshua

PR #411, opened 2019-03-04
Task/DES-684 Modified location of Data Depot header, moved the 'Learn About The Da…

eliasmarkc:
Created data depot header on top of left nav, moved toolbar to top of ui-view.
Added Learn about Data Depot link under Curation Guidelines link.
Fixed wrapping issue with toolbar buttons, but justifying the "Find in..." text box to the left and the buttons to the right.
screen shot 2019-03-04 at 10 12 02 am
screen shot 2019-03-04 at 10 12 21 am

### DIFF
--- a/designsafe/static/scripts/data-depot/components/data-depot-toolbar/data-depot-toolbar.component.html
+++ b/designsafe/static/scripts/data-depot/components/data-depot-toolbar/data-depot-toolbar.component.html
@@ -10,82 +10,84 @@
             </form>
             -->
         </div>
-        <div class="btn-toolbar btn-toolbar-right" role="toolbar" aria-label="Data Browser Toolbar">
+        <div class="btn-toolbar btn-toolbar-left" role="toolbar" aria-label="Data Browser Toolbar">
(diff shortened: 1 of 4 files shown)


designsafe/static/scripts/data-depot/components/data-depot-toolbar/data-depot-toolbar.component.html
```diff
@@ -16,7 +16,7 @@
                 <input class="form-control" style="width:240px;" placeholder="Find in {{ $ctrl.placeholder() }}" ng-model="$ctrl.search.queryString">
                 <button class="btn btn-default" style="clear: right;"> <i class="fa fa-search"></i> </button>
             </form>
-            <div class="btn-group">
+            <div class="btn-group" style="margin-left:0px">
```
xirdneh: Can you move this to a CSS class?
We're trying to move away from inline styles.

eliasmarkc: Done.

PR #412, opened 2019-03-04
task/DES-752 Create alert system

eliasmarkc:
Alerts loaded into session on agave callback post login
Page alert model created
Added front end and styling
Test added


designsafe/apps/api/messages/models.py
```diff
@@ -0,0 +1,21 @@
...
+    ('alert-success', 'SUCCESS'),
+    )
+    
+class PageAlert(models.Model):
+    alert_id = models.CharField(max_length=120, unique=True, primary_key=True)
+    alert_type = models.CharField(max_length=20, choices=TYPE_CHOICES, default='alert-warning')
+    alert_message = models.TextField(help_text='Page alert message text.')
+    
+    def __str__(self):
+        return unicode(self).encode('utf-8')
```
xirdneh: The `__str__` method needs to be more explicit.
```suggestion
        return '{alert_id}: {alert_type} - {alert_message}...'.format(
            alert_id=self.alert_id,
            alert_type=self.alert_type,
            alert_message=self.alert_message[0:50]
        )```

File: designsafe/apps/api/messages/models.py [hunk 2]
```diff
@@ -0,0 +1,21 @@
+import logging
```
xirdneh: Add `import unicode_literals`: https://python-future.org/unicode_literals.html

designsafe/apps/api/messages/models.py
```diff
@@ -0,0 +1,21 @@
...
+    
+class PageAlert(models.Model):
+    alert_id = models.CharField(max_length=120, unique=True, primary_key=True)
+    alert_type = models.CharField(max_length=20, choices=TYPE_CHOICES, default='alert-warning')
+    alert_message = models.TextField(help_text='Page alert message text.')
+    
+    def __str__(self):
+        return unicode(self).encode('utf-8')
+
+    def __unicode__(self):
```
xirdneh: `__str__` is enough. It would be better to also add `__repr__`: https://python-reference.readthedocs.io/en/latest/docs/functions/repr.html

eliasmarkc:
> Why do we need to save the alerts in the database? Django's alert framework already saves the alerts in the session. Do we have to persist those?

These alerts are supposed to persist and take less space in the browser. This was listed in part of the requirements. The respective Jira task has information on this.

designsafe/apps/api/messages/models.py
eliasmarkc: Importing this broke everything.

xirdneh: Why did it returned a 502? What was the actual error?

designsafe/apps/api/messages/tests.py
```diff
@@ -0,0 +1,39 @@
+from designsafe.apps.api.messages.models import PageAlert
+from designsafe.apps.auth.views import load_alerts_into_session, agave_oauth_callback
+from django.test import TestCase
+from django.test.client import RequestFactory
+from django.forms.models import model_to_dict
+from django.contrib.sessions.middleware import SessionMiddleware
+from django.contrib.messages.storage.fallback import FallbackStorage
+from mock import patch, Mock
+
+alerts = []
```
xirdneh: Not ideal to use a model level constant. This can be done on `setUp` or `setUpTestData` and attach it to the class.

designsafe/apps/api/messages/tests.py
```diff
@@ -0,0 +1,39 @@
...
+        request = RequestFactory().get('/')
+        add_session_to_request(request)
+        load_alerts_into_session(request)
+        alert_messages = []
+        for alert in alerts:
+            alert_messages.append(model_to_dict(alert))
+        self.assertEqual(request.session['alertslist'], alert_messages)
+        self.assertEqual(request.session['alertslist_size'], len(alert_messages))
+
+def add_session_to_request(request):
```
xirdneh: Not ideal to have a model level function like this. It's ok if unit tests are verbose. Writing unit test is not the same as writing application code. You can move this inside the test.

designsafe/apps/api/messages/models.py
eliasmarkc: No idea, I tried to reproduce it and it seemed to have worked out okay.

designsafe/apps/api/messages/tests.py
eliasmarkc: Fixed.

designsafe/apps/api/messages/tests.py
eliasmarkc: Fixed.

designsafe/apps/api/messages/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+logger = logging.getLogger(__name__)
+TYPE_CHOICES = (
+    ('alert-warning', 'WARNING'),
+    ('alert-danger', 'DANGER'),
+    ('alert-info', 'INFO'),
+    ('alert-success', 'SUCCESS'),
+    )
+
+class PageAlert(models.Model):
```
xirdneh: @rstijerina Do we want to add a `is_active` attribute to keep a history of past alerts and not having to delete them if we don't want to show them?

designsafe/apps/auth/views.py
```diff
@@ -198,3 +203,9 @@ def agave_oauth_callback(request):
 
 def agave_session_error(request):
     return render(request, 'designsafe/apps/auth/agave_session_error.html')
+
+# load_alerts_into_session() will be called after a new login ( within agave_oauth_callback() )
+def load_alerts_into_session(request):
+    alert_messages = [model_to_dict(alert) for alert in PageAlert.objects.all()]
```
xirdneh: Doesn't `model_to_dict` also adds the model type and id to the returned object?
Do we need that extra info there?
Maybe we could just do something more granular like:
```suggestion
    alert_messages = [{"alert_type": alert.alert_type, "alert_message": alert.alert_message} for alert in PageAlert.objects.all()]
```

designsafe/apps/auth/views.py
```diff
@@ -107,8 +110,10 @@ def agave_oauth_callback(request):
     """
     http://agaveapi.co/documentation/authorization-guide/#authorization_code_flow
     """
+    load_alerts_into_session(request)
```
xirdneh: I think this would probably need to be in a middleware. If somebody creates a new alert when a user is already logged in that user is not going to see this.
If we put it in a middleware then we can add the alerts to the session every time a user does a request to the user.
Then we find another issue, if that's a request within the data depo the user is not going to see it unless they fully refresh the page.
We could take care of this but I think that's out of this scope. For now we can move this to a middle ware and then create another ticket to see how can we show alerts within a section in the portal that only does partial updates.

designsafe/apps/api/messages/models.py
```diff
@@ -0,0 +1,30 @@
...
+    alert_id = models.CharField(max_length=120, unique=True, primary_key=True)
```
xirdneh: Why specifying a different field for the Id? By default Django uses `id` as a primary key.
Is this supposed to be filled by the admin user?

designsafe/apps/api/messages/models.py
```diff
@@ -0,0 +1,30 @@
...
+    alert_type = models.CharField(max_length=20, choices=TYPE_CHOICES, default='alert-warning')
```
xirdneh: Do we need to prepend each field with `alert_` the model class is already named `PageAlert` we could name them like:
```python
class PageAler(models.Model):
    type = models.CharField(max_length=20)
    message = models.TextField(help_text='Page alert message text.')
```

designsafe/apps/auth/views.py
eliasmarkc: The instructions were to only pull alerts on login. I brought up the same issues you brought up with this and learned that is expected behavior.

designsafe/apps/api/messages/models.py
eliasmarkc: Right, it's supposed to be filled out by an admin. Should the field be renamed to `name`?

designsafe/apps/api/messages/models.py
eliasmarkc: I'll fix that

designsafe/apps/api/messages/models.py
rstijerina: Yeah that might be a good idea, especially if we find we have a few standard alerts that we'll need to activate periodically for hpc downtimes, etc.

PR #424, opened 2019-03-12
[WIP] DES-749 Built Views for Django

climberfields:
Built views for GET POST and Delete requests, submitting a PR as a [WIP] per Josue


designsafe/apps/api/notifications/tests.py
```diff
@@ -176,3 +176,58 @@ def test_webhook_interactive_post(self, get_user_model, mock_receiver):
         self.assertEqual(action_link, link_from_event)
         self.assertEqual(n.operation, 'web_link')
         self.assertEqual(response.status_code, 200)
+class NotificationsViewsTests(TestCase):
```
xirdneh: Double space before a `class` please. Check your code with `flake8` and `pylint` if possible.

designsafe/apps/api/notifications/tests.py
```diff
@@ -176,3 +176,58 @@ def test_webhook_interactive_post(self, get_user_model, mock_receiver):
...
+    def setUp(self):
+        #user = get_user_model().objects.get(username="ds_user")
+        note = Notification(
+            event_type="JOB",
+            status="SUCCESS",
+            message="Job succeeded",
+            user="ds_user",
+        )
+        note.save()
+        #self.client=Client()
```
xirdneh: Try to avoid having commented code on the finished version. If the code is already in a past commit you can hunt it back by looking at the history of the file.

designsafe/apps/data/views/base.py
```diff
@@ -174,3 +175,33 @@ def get(self, request, file_mgr_name, system_id, file_path):
...
+    date = WordView.datetime.datetime
+
+
+    def get(self):
+        term = required(self.input())
+        definition = required(input())
+        vocabBuilder = logger.logging.getLogger(__name__)
+        date
+        permalink = required(input())
+
```
xirdneh: Only one space between class methods, please.

designsafe/apps/data/views/base.py
```diff
@@ -174,3 +175,33 @@ def get(self, request, file_mgr_name, system_id, file_path):
...
+
+
+    def post(self):
+        vocabBuilder = logger.logging.getLogger(__name__)
```
xirdneh: Define logger at the module level, please.

designsafe/apps/data/views/base.py
```diff
@@ -174,3 +175,33 @@ def get(self, request, file_mgr_name, system_id, file_path):
...
+        term = required(self.input())
+        definition = required(input())
+        permalink = required(input())
+
+
+
+    
+    def delete(self):
+        delete = self.client.del
```
xirdneh: Not sure what this is doing?

es_setup.sh
```diff
@@ -0,0 +1,126 @@
+npm install --save-dev elasticdump
```
xirdneh: What are we trying to do here?

climberfields: I'm not sure where this came from. I had not typed this. Maybe it's the result of a different pulldown?

SilversunKSauri: This is Jake's elasticsearch dump script.  Only needed for local dev.  You probably put it in your designsafe code folder and accidentally tracked it with git.

designsafe/apps/data/models/taxonomy.py
```diff
@@ -0,0 +1,21 @@
+from datetime import datetime
+from django.db import models
+from django.contrib.auth.models import User
+from django.contrib.auth.validators import ASCIIUsernameValidator
+
+
+
+
+class CustomUser(User):
```
xirdneh: Keep it to two spaces before a class definition.
Also, here you are creating a new model based on `User`. This is not needed, if you want to have a user bound to each record in `Vocab` you can do it like so:
```python
user = models.ForeignKey(settings.AUTH_USER_MODEL)
```

designsafe/apps/data/models/taxonomy.py
```diff
@@ -0,0 +1,21 @@
...
+    username_validator = ASCIIUsernameValidator()
+
+class Vocab(models.Model):
+    Term = models.CharField(max_length=40, unique=True)
```
xirdneh: Class attributes have to be PEP8 compliant. i.e. `term`, `definition`, `source_of_definition`, `date` and `vocab_builder`

designsafe/apps/data/models/taxonomy.py
```diff
@@ -0,0 +1,21 @@
...
+    Definition = models.TextField(max_length=200)
+    Source_of_Definition = models.UrlField(max_length=1000, unique=True)
+    Date = DateTimeField(default=datetime.now(), auto_now_add=True)
+    VocabBuilder = CustomUser()
```
xirdneh: If this is to bind a user to a vocab record then use `ForeignKey`

climberfields: @xirdneh use ForeignKey for VocabBuilder? Or just for Vocab in general? The VocabBuilder feature is to list the individual that created the vocabulary word in the system.

xirdneh: For `VocabBuilder` Something like:
```suggestion
    VocabBuilder = models.ForeignKey(settings.AUTH_USER_MODEL)
```
You might want to check the previous syntax I just typed it by memory and I usually make mistakes :)
You can check the [docs](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey) for more info.
Also, we're already doing something like that in other [models](https://github.com/DesignSafe-CI/portal/blob/master/designsafe/apps/auth/models.py#L20).

climberfields: Gotcha, that's what I was thinking, I just wanted to make sure. Thanks for your help!

climberfields:
Changes made

designsafe/apps/api/notifications/tests.py
```diff
@@ -30,7 +30,7 @@
 
 # Create your tests here.
 @skip("Need to mock websocket call to redis")
-class NotificationsTestCase(TestCase):
+  class NotificationsTestCase(TestCase):
```
rstijerina: Classes shouldn't be indented like this, I believe Josue was suggesting to keep two line breaks before the first class definition in a given file

```suggestion
class NotificationsTestCase(TestCase):
```

designsafe/apps/data/views/base.py
```diff
@@ -181,6 +181,7 @@ def get(self, request, file_mgr_name, system_id, file_path):
         return response
 
 
+
```
rstijerina: You can remove this extra line
```suggestion
```

PR #425, opened 2019-03-13
Task/des 155: New UI for publications listing/search.

jarosenb:
- What is the status of the PR?
[X ] Ready.
[] Work in Progress.
[ ] Hold.
- Are there any migrations?
[ ] Yes.
[X] No.
- Description:
Implements Craig's new UI for search in Published Projects
- Summary of Changes:
Formatting changes and new columns for the publication listing view. Checkbox selection for different publication types. 
pubs_new_ui


designsafe/apps/api/agave/filemanager/public_search_index.py
```diff
@@ -26,7 +26,7 @@
 
 class PublicationIndexed(DocType):
     class Meta:
-        index = settings.ES_INDICES['publications']['name']
+        index = settings.ES_INDICES['publications']['alias'][0]
```
xirdneh: Can we avoid hard coding a list index?

designsafe/apps/api/agave/filemanager/public_search_index.py
```diff
@@ -444,7 +446,9 @@ def search(self, system, query_string,
         nees_published_res = nees_published_search.execute()
         des_published_res = des_published_search.execute()
 
-        logger.debug(nees_published_res.hits.total)
+        for res in des_published_res:
+            new_desc = re.sub(query_string, lambda x: '<b>{}</b>'.format(x.group(0)), res.project.value.description)
```
xirdneh: Doesn't elasticsearch automatically highlights the query string on the result?

designsafe/static/scripts/data-depot/components/dd-publication-listing/dd-publication-listing.component.js
```diff
@@ -1,5 +1,11 @@
+class ddPublicationListingController {
+    constructor($stateParams) {
```
xirdneh: add `'ngInject';`

designsafe/apps/api/agave/filemanager/public_search_index.py
jarosenb: Would it be OK to tweak `settings.ES_INDICES` to be a string instead of a list? AFAIK we've never had a case where we've actually wanted multiple aliases for one index.

designsafe/apps/api/agave/filemanager/public_search_index.py
jarosenb: Elasticsearch has a highlighting option, but it would require reindexing `des-publications` with Object fields instead of Nested fields. The highlights are returned separately from the main search results, so that would all still have to be stitched together to display on the front end.

designsafe/static/scripts/data-depot/components/dd-publication-listing/dd-publication-listing.component.js
jarosenb: Done.

designsafe/apps/api/agave/filemanager/public_search_index.py
xirdneh: Yeah, that's ok.

jarosenb: Cool, I pushed up that change.

designsafe/apps/api/agave/filemanager/public_search_index.py
xirdneh: Even if we use something [like `inner_hits`](https://www.elastic.co/guide/en/elasticsearch/reference/5.3/nested.html#CO181-4) ?
I think it's ok if the highlights are in a separate object, we can handle that on the frontend.
The thing is that if we use ES highlighter we get two things:
1. We don't tell the frontend how things are going to be rendered from the backend. Instead, we send enough information so the frontend makes the decision about how things are going to be rendered.
2. Depending on the analyzers the `query_string` is not always going to be exactly what's going to be the actual hit on the search. So using ES's highlighter would make that more accurate.

If this is a show stopper we can move ahead with this and create a ticket to update it to use ES highlighter.

jarosenb: I think updating the highlighter should be a separate ticket and include reindexing `des-publications` to get rid of those nested fields. If we wanted highlights with the current index it would involve rewriting our basic search query as 3 or 4 separate multi-layer nested queries depending on what fields we want to hit. Once we reindex (which will be pretty trivial now that the swap reindexer has been merged in) we'll just be able to pass a list of fields and get highlights without having to worry about the nested structure.

jarosenb:
Putting a WIP label on this until I get it working with the new data depot components.

designsafe/apps/api/agave/filemanager/public_search_index.py
xirdneh: @jarosenb We have to go over this implementation.
Having parent/children relation between objects is not always the best way to go.

jarosenb: I agree, a lot of our `Nested` fields make sense to have that way, but there are some (like having a publication's title/description nested 3 levels deep) that seem to make life harder for no reason.

xirdneh: We do need to change the structure of publications. But this is planned for after May.

PR #459, opened 2019-04-05
Task/DES-656 Zip published project data with celery task

Strmiska:
Creates a zip file within the published project's directory containing the project's files.

### DIFF
--- a/designsafe/apps/api/agave/filemanager/public_search_index.py
+++ b/designsafe/apps/api/agave/filemanager/public_search_index.py
@@ -13,6 +13,9 @@ import os
 import re
 import datetime
 import itertools
+import zipfile
+import shutil
+import os
 from django.conf import settings
 from elasticsearch import TransportError, ConnectionTimeout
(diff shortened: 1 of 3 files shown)


xirdneh:
@Strmiska Can we move the implementation to the [project class](https://github.com/DesignSafe-CI/portal/blob/master/designsafe/apps/projects/models/agave/base.py)?
That way everything related to the project is there and we don't have such long tasks.
Also, that way we can add it to a django management command easily and use it in other places.

Strmiska:
> @Strmiska Can we move the implementation to the [project class](https://github.com/DesignSafe-CI/portal/blob/master/designsafe/apps/projects/models/agave/base.py)?
> That way everything related to the project is there and we don't have such long tasks.
> Also, that way we can add it to a django management command easily and use it in other places.

@xirdneh I've moved **archive** into the **project** and **published project** classes. I'm not sure if this is redundant, but I thought the alternative to placing it in one and importing the other class seemed odd. What do you think?

PR #466, opened 2019-04-17
Bug/DES-953 - Fix issue with team members not being able to access projects

Strmiska:
Users added onto a project would be added to project metadata but would not have access to the project in their project directory.


Strmiska:
I thought there was a way we could improve setting the project's related entity meta permissions when updating the project form, but after digging through it I'm not sure that is the case anymore.

PR #468, opened 2019-04-22
Task/des 795

fnets:
- What is the status of the PR?

- [ ] Ready.
- [X] Work in Progress.
- [ ] Hold.

- Are there any migrations?

- [ ] Yes.
- [X] No.

- Description:

 Removing datepicker from date field on edit experiment form, so that dates do not have to be re-entered when editing an experiment.


fnets:
This PR is no longer needed

PR #67, opened 2017-11-17
Add highlighting to My Projects page in Data Depot

MattShirley:
This adds functionality to My Projects to highlight project, similar to My Data. Currently, clicking anywhere on the project row will progress to the Project Detail page.

This adds a selectedProjects property to the currentState object in Data Browser Service. There are also methods called selectProjects, deselectProjects, and toggleProjects to manage this property. This works similarly to the selected files property. It doesn't require the input to be a list of projects (it can be a single project).

The ProjectListingCtrl controller is augmented to use these new Data Browser Service methods. It supports functionality similar to the My Data page. Single projects can be accessed with a click, multiple projects can be selected by holding the Command/Control button, and multiple projects can be selected by holding the shift key.

PR #487, opened 2019-05-08
task/DES-1006: Don't show files in other users' My Data

jarosenb:
Before, if you tried to access another user's My Data, we would check that user's files for anything you had Read permission for and display those. This is unintuitive/potentially insecure, so now My Data is locked to a user's own files.

### DIFF
--- a/designsafe/apps/api/agave/views.py
+++ b/designsafe/apps/api/agave/views.py
@@ -63,8 +63,7 @@ class FileListingView(BaseApiView):
                 file_path = request.user.username
 
             if system_id == AgaveFileManager.DEFAULT_SYSTEM_ID and \
-                (file_path.strip('/') == '$SHARE' or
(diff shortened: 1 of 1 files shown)


designsafe/apps/api/agave/views.py
```diff
@@ -61,15 +61,6 @@ def get(self, request, file_mgr_name, system_id=None, file_path=None):
                 system_id = AgaveFileManager.DEFAULT_SYSTEM_ID
             if file_path is None:
                 file_path = request.user.username
-
-            if system_id == AgaveFileManager.DEFAULT_SYSTEM_ID and \
-                (file_path.strip('/') == '$SHARE' or
-                 file_path.strip('/').split('/')[0] != request.user.username):
```
rstijerina: Maybe just remove this line #67?

PR #509, opened 2019-05-17
Bug/DES-1061 - Fixing issues related to user permissions being set on projects

Strmiska:


designsafe/apps/api/projects/views.py
```diff
@@ -304,17 +304,19 @@ def post(self, request, project_id):
...
         new_pi = post_data.get('pi', p.pi)
-        new_co_pis = post_data.get('copi', p.co_pis)
+        new_co_pis = post_data.get('coPis', p.co_pis)
         new_team_members = post_data.get('teamMembers', p.team_members)
 
-        if new_pi and new_pi != 'null' and p.pi != new_pi:
-            names = list(set(p.team_members + p.co_pis + [p.pi]))
+        # we need to compare the existing project data with the updated project data
+        if new_pi and new_pi != 'null' and meta_obj.value['pi'] != new_pi:
+            names = list(set(meta_obj.value['teamMembers'] + meta_obj.value['coPis'] + [meta_obj.value['pi']]))
```
rstijerina: `p` here is instantiated with `meta_obj`, so I don't think this is necessary

PR #518, opened 2019-05-21
Formatting; Fixes forEach bug in pipeline

rstijerina:


designsafe/apps/api/tasks.py
```diff
@@ -996,12 +996,13 @@ def email_collaborator_added_to_project(self, project_title, team_members_to_add
...
                 logger.debug('this is the user:'.format(collab_user))
                 try:
-                    collab_user.profile.send_mail("You have been added to a DesignSafe project!", '''\
-                        Hi {},
-                        You have been added to the project {}.
-                        You can now start working on the project. Please use your TACC account to access the DesignSafe-CI website or to ask for help.
-                        <This is a programmatically generated message. Do NOT reply to this message.>
-                        Thanks, The DesignSafe-CI team"\
-                        '''.format(collab_user.get_full_name(), project_title))
+                    collab_user.profile.send_mail("You have been added to a DesignSafe project!", """
```
SilversunKSauri: Wasn't this part of a different PR that got merged today (5/23)?

rstijerina: ah yes it was, thanks

fnets:
Order of authors was preserved.  I can't open the Citation on the experiment during the publication preview.  Looking through the Data Browser Service, some of this gets pulled from the DOI but some don't.   I think we should make it more clear that the button is disabled if we don't expect it to work.

rstijerina:
Keith said it looks good :)

rstijerina:
> Order of authors was preserved. I can't open the Citation on the experiment during the publication preview. Looking through the Data Browser Service, some of this gets pulled from the DOI but some don't. I think we should make it more clear that the button is disabled if we don't expect it to work.

@fnets Can you make a task for that?

PR #523, opened 2019-05-22
Bug/DES-1075 - Add published details to other template

Strmiska:


jarosenb:
Works for me locally.

PR #525, opened 2019-05-23
Fix 'None' str publication end date

rstijerina:
From the stacktrace it seems as though elasticsearch does not like `publicationEnd` to be `'None'` as a string. This should fix that, but it's kinda ugly. LMK if you have a better idea for a solution.

### DIFF
--- a/designsafe/apps/api/agave/filemanager/public_search_index.py
+++ b/designsafe/apps/api/agave/filemanager/public_search_index.py
@@ -591,10 +591,13 @@ class PublicationManager(object):
         publication['licenses'] = publication.pop('license', [])
         publication['license'] = ''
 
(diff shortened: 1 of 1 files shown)


jarosenb:
I think a prettier solution might be to make sure that `procedureEnd` is always either a date or `null` before we send it to the back end (I don't know why `manage-experiments.component` line 213 isn't preventing this already). This option looks good enough for now, though.

rstijerina:
@jarosenb Yeah, that was my initial thought, but on the back end the model enforces the field to a custom CharField. So it transforms `null` to `'None'`, which gives us these errors. My other thought for a solution was to make the field `procedure_end` on the `Simulation` model to accept `None` types

PR #556, opened 2019-06-03
Task/DES-1111 - Change file apiParams for published area

Strmiska:
It looks like files were redirecting and 404ing in the published area because they did not have the expected apiParams.

### DIFF
--- a/designsafe/static/scripts/projects/components/file-categories/file-categories.component.js
+++ b/designsafe/static/scripts/projects/components/file-categories/file-categories.component.js
@@ -10,6 +10,10 @@ import { values } from '@uirouter/core';
 
 const getFileUuid = (file) => {
     let promise;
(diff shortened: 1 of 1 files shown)


designsafe/static/scripts/projects/components/file-categories/file-categories.component.js
```diff
@@ -10,6 +10,10 @@ import { values } from '@uirouter/core';
 
 const getFileUuid = (file) => {
     let promise;
+    if (file.apiParams.fileMgr === 'published') {
```
rstijerina: Could this be handled in the state provider?
https://github.com/DesignSafe-CI/portal/blob/87bfcb787b82de2697228b853bebb4cf0fc2276e/designsafe/static/scripts/data-depot/index.js#L951

PR #558, opened 2019-06-03
Hotfix-Bug/DES-1113 - Change archive path for publication

Strmiska:
- The `zip_publication_files` task was pointed to the wrong directory. This caused permissions issues with projects.
- Commented out `zip_project_files` since it has not been finished.
- This fixes the issue of permissions being changed on the projects directory instead of the publications directory


reptillicus:
Make JIRA ticket for adding try/except for permissions fixin'

PR #559, opened 2019-06-05
Task/DES-1115 - Consolidate requests for user data where possible

Strmiska:
- Updated directives meant to format usernames for project related views.
- Condensed individual requests for user data in `move/copy` modals and in `project-listing`
- Updated templates where applicable

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
+++ b/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
@@ -92,18 +92,14 @@
             </colgroup>
             <tr class="prj-row">
                 <td>PI</td>
-                <td>
-                    <ds-user class="prj-data" username="$ctrl.project.value.pi" format="lname">{{ $ctrl.project.value.pi }}
-                    </ds-user>
+                <td class="prj-data">
+                    <ds-user-list usernames="[$ctrl.project.value.pi]"></ds-user-list>
                 </td>
             </tr>
             <tr ng-if="$ctrl.project.value.coPis.length">
                 <td>CoPIs</td>
                 <td class="prj-data">
-                    <span ng-repeat="copi in $ctrl.project.value.coPis">
-                        <span ng-if="!$last"><ds-user-list usernames="copi"></ds-user-list>; </span>
-                        <span ng-if="$last"><ds-user-list usernames="copi"></ds-user-list></span>
-                    </span>
+                    <ds-user-list usernames="$ctrl.project.value.coPis"></ds-user-list>
                 </td>
             </tr>
             <tr class="prj-row" ng-hide="$ctrl.project.value.projectType === 'other' && $ctrl.project.value.dataType">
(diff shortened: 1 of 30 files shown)


designsafe/static/scripts/data-depot/components/projects/project-listing/project-listing.component.js
```diff
@@ -18,9 +19,27 @@ function ProjectListingCtrl($scope, $state, DataBrowserService, Django, ProjectS
     ProjectService.list({offset:offset, limit:limit}).then(function(projects) {
       $scope.ui.busy = false;
       $scope.data.projects = _.map(projects, function(p) { p.href = $state.href('projects.view', {projectId: p.uuid}); return p; });
-      DataBrowserService.projectBreadcrumbSubject.next()
+      DataBrowserService.projectBreadcrumbSubject.next();
+      $scope.getNames();
     });
 
+    $scope.getNames = function () {
```
rstijerina: I think this function should send over an array of usernames as a payload to the `getPublic` call. This may require some rewriting, but it is more extensible than using separators.

Strmiska: That shouldn't be a big change though. I'll try passing the usernames array as something like `JSON.stringify(usernames)`? I can poke around on it tomorrow.

designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
```diff
@@ -92,18 +92,14 @@ <h3 style="text-align: center; color: orange;">
             </colgroup>
             <tr class="prj-row">
                 <td>PI</td>
-                <td>
-                    <ds-user class="prj-data" username="$ctrl.project.value.pi" format="lname">{{ $ctrl.project.value.pi }}
-                    </ds-user>
+                <td class="prj-data">
+                    <ds-user-list usernames="[$ctrl.project.value.pi]"></ds-user-list>
```
SilversunKSauri: Just curious, why is this in [] when other similar calls (like the ds-user-list below this on line 102) don't have it?

designsafe/static/scripts/ng-designsafe/services/user-service.js
```diff
@@ -15,7 +15,7 @@ export function UserService($http, $q) {
     };
 
     this.getPublic = function (username) {
-      return $http.get("/api/users/public", {params: {username: username}})
+      return $http.get("/api/users/public/", {params: {username: JSON.stringify(username)}})
```
rstijerina: should we change references here to `usernames` instead of `username` if we're always assuming an array now?

designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
Strmiska: At the moment we only have one pi on a project, and their username is in string format. Co-PIs and Team Members are in arrays of string usernames. I wrap the PI in brackets here so that the `dsUserList` directive will be able to iterate over it.

designsafe/static/scripts/ng-designsafe/services/user-service.js
Strmiska: Yes that makes sense. I'll swap it out.

Strmiska:
Pushed requested changes

PR #565, opened 2019-06-13
bug/DES-1124: Fix bad field validation; Other App Form fixes

rstijerina:
- Fixes processorsPerNode calculation.
- Accommodate missing defaults in app form.
- Overhauls maxRunTime field in app job submission form. Now strictly validates based on defaultMaxRunTime defined in the app, or the maxRequestedTime of the queue if no default given.
- Fixes a bug caused by blank inputs by removing them from the job submission.
- Suppresses uncaught exceptions on client side qProvider

### DIFF
--- a/designsafe/static/scripts/workspace/controllers/application-form.js
+++ b/designsafe/static/scripts/workspace/controllers/application-form.js
@@ -207,7 +207,7 @@ export default function ApplicationFormCtrl($scope, $rootScope, $localStorage, $
             } else if (('nodeCount' in jobData) && ('processorsPerNode' in jobData)) {
                 jobData.processorsPerNode = jobData.nodeCount * jobData.processorsPerNode;
             } else if (!('nodeCount' in jobData) && ('processorsPerNode' in jobData)) {
-                jobData.processorsPerNode = jobData.defaultNodeCount * jobData.processorsPerNode;
+                jobData.processorsPerNode = $scope.data.app.defaultNodeCount * jobData.processorsPerNode;
             }
 
             $scope.jobReady = true;
--- a/designsafe/static/scripts/workspace/controllers/application-form.js
+++ b/designsafe/static/scripts/workspace/controllers/application-form.js
@@ -203,11 +203,11 @@ export default function ApplicationFormCtrl($scope, $rootScope, $localStorage, $
 
             // Calculate processorsPerNode if nodeCount parameter submitted
             if (('nodeCount' in jobData) && !('processorsPerNode' in jobData)) {
-                jobData.processorsPerNode = jobData.nodeCount * ($scope.data.app.defaultProcessorsPerNode / $scope.data.app.defaultNodeCount);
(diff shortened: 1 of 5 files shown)


designsafe/static/scripts/workspace/controllers/application-form.js
```diff
@@ -203,11 +203,11 @@ export default function ApplicationFormCtrl($scope, $rootScope, $localStorage, $
 
             // Calculate processorsPerNode if nodeCount parameter submitted
             if (('nodeCount' in jobData) && !('processorsPerNode' in jobData)) {
-                jobData.processorsPerNode = jobData.nodeCount * ($scope.data.app.defaultProcessorsPerNode / $scope.data.app.defaultNodeCount);
+                jobData.processorsPerNode = jobData.nodeCount * ($scope.data.app.defaultProcessorsPerNode || 1) / ($scope.data.app.defaultNodeCount || 1);
```
Strmiska: Does this need another set of parenthesis around `($scope.data.app.defaultProcessorsPerNode || 1) / ($scope.data.app.defaultNodeCount || 1)`?

rstijerina: Nah, that extra set of parenthesis was redundant here since any order of operations will yield the same result

designsafe/static/scripts/workspace/services/apps-service.js
```diff
@@ -217,11 +276,11 @@ export function appsService($http, $q, $translate, djangoUrl, Django) {
...
             title: 'Processors Per Node',
             description: `Number of processors (cores) per node for the job. e.g. A selection of 16 processors per node along with 4 nodes
-            will result in 16 processors on 4 nodes, with 64 processors total. Default number of processors per node is ${Math.floor(app.defaultProcessorsPerNode / app.defaultNodeCount)}.`,
+            will result in 4 nodes with 16 processors each, 64 processors total. Default number of processors per node is ${Math.floor(app.defaultProcessorsPerNode || 1) / (app.defaultNodeCount || 1)}.`,
             type: 'integer',
-            default: Math.floor(app.defaultProcessorsPerNode / app.defaultNodeCount),
+            default: Math.floor((app.defaultProcessorsPerNode || 1) / (app.defaultNodeCount || 1)),
             minimum: 1,
-            maximum: Math.floor(app.defaultProcessorsPerNode / app.defaultNodeCount),
+            maximum: Math.floor(app.defaultProcessorsPerNode || 1) / (app.defaultNodeCount || 1),
```
Strmiska: Does `maximum` only need `app.defaultProcessorsPerNode` to be rounded or should it look like `default` too?

rstijerina: It needs to look like default, since in the definition `defaultProcessorsPerNode` is actually a total of all processors across all nodes. So here we break it down into a value that actually represents `processorsPerNode`

PR #569, opened 2019-06-19
bug/DES-1124: Remove falsy inputs/parameters for aloe

rstijerina:


designsafe/static/scripts/workspace/services/apps-service.js
```diff
@@ -219,7 +219,7 @@ export function appsService($http, $q, $translate, djangoUrl, Django) {
                     tmp = replaceAt(tmp, index, n);
                     if (Object.is(arr.length - 1, i) || (arr[i + 1] == 0)) {
```
jarosenb: Not specifically related to this PR, but what's the reasoning behind using `Object.is` when comparing 2 numbers in this function?

rstijerina: Good point, I think before I was looking at indices but left `Object.is` in there. I'll switch it out

PR #570, opened 2019-06-27
Bug/DES-1144 - Fix user directive to support other project authors (hotfix)

Strmiska:
Created a condition inside the `dsUserList` directive to accept a list of user objects for other
Added try/except catch if user's name is not found in database

### DIFF
--- a/designsafe/apps/api/agave/filemanager/public_search_index.py
+++ b/designsafe/apps/api/agave/filemanager/public_search_index.py
@@ -145,11 +145,11 @@ class Publication(object):
                 last_name=pi_user['last_name'], first_name=pi_user['first_name'])
         else:
             try:
-                pi_username = get_user_model().objects.get(username=self.project['value']['pi'])
+                pi_user = get_user_model().objects.get(username=pi)
                 dict_obj['meta']['piLabel'] = '{last_name}, {first_name}'.format(
-                    last_name=pi_username.last_name, first_name=pi_username.first_name)
+                    last_name=pi_user.last_name, first_name=pi_user.first_name)
             except:
-                dict_obj['meta']['piLabel'] = pi
+                dict_obj['meta']['piLabel'] = '({pi})'.format(pi=pi)
 
         return dict_obj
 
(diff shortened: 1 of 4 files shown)


designsafe/apps/api/agave/filemanager/public_search_index.py
```diff
@@ -144,9 +144,13 @@ def to_file(self):
             dict_obj['meta']['piLabel'] = '{last_name}, {first_name}'.format(
                 last_name=pi_user['last_name'], first_name=pi_user['first_name'])
         else:
-            pi_username = get_user_model().objects.get(username=self.project['value']['pi'])
-            dict_obj['meta']['piLabel'] = '{last_name}, {first_name}'.format(
-                last_name=pi_username.last_name, first_name=pi_username.first_name)
+            try:
+                pi_username = get_user_model().objects.get(username=self.project['value']['pi'])
```
rstijerina: if you have the `pi` object defined above on line 140, why not use it for all instances of `self.project['value']['pi']`?

Also the `pi_username` object is really a User instance, not a username, so `pi_username` might be `pi_user` instead

designsafe/static/scripts/ng-designsafe/directives/ng-designsafe-directives.js
```diff
@@ -277,30 +277,60 @@ export function dsUserList(UserService) {
...
+            });
+            element.text(formattedNames);
           });
-          element.text(formattedNames);
-        });
+
+        } else {
+          UserService.getPublic(userReq).then((res) => {
+            let userData = res.userData;
+            let formattedNames = '';
```
rstijerina: some of this is repeated above - should it be outside the if statement?

designsafe/apps/api/agave/filemanager/public_search_index.py
Strmiska: Good call. I missed the other `pi` references there, and on the `pi_username` yes that makes more sense I'll change that to `pi_user`.

designsafe/static/scripts/ng-designsafe/directives/ng-designsafe-directives.js
Strmiska: Yep! I'll move it out.

PR #571, opened 2019-06-27
Bug/DES-1145 - Order other authors in xml sent to DataCite

Strmiska:
Sort authors by `order` for DataCite XML

### DIFF
--- a/designsafe/apps/api/projects/managers/publication.py
+++ b/designsafe/apps/api/projects/managers/publication.py
@@ -166,13 +166,36 @@ def _project_required_xml(project, authors, created, doi=None):
         identifier.text = SHOULDER.replace('doi:', '')
     creators = ET.SubElement(resource, 'creators')
 
-    for author in authors:
-        if not author.get('lname') and not author.get('fname'):
-            logger.error('Author in wrong format: %s', author)
-            continue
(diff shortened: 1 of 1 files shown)


designsafe/apps/api/projects/managers/publication.py
```diff
@@ -166,13 +166,36 @@ def _project_required_xml(project, authors, created, doi=None):
...
     creators = ET.SubElement(resource, 'creators')
 
-    for author in authors:
-        if not author.get('lname') and not author.get('fname'):
-            logger.error('Author in wrong format: %s', author)
-            continue
-        creator = ET.SubElement(creators, 'creator')
-        creator_name = ET.SubElement(creator, 'creatorName')
-        creator_name.text = '{}, {}'.format(author['lname'], author['fname'])
+    if project.value.projectType == 'other':
```
Strmiska: Having difficulty tracking down the data type of `project` without actually publishing a project. I believe it is formatted as a request, but I'm not sure. The ordering of authors on the XML is working, but I'm not sure if `project.value.projectType` is going to work in this case.

xirdneh:
We can probably close this PR because of the recent Datacite changes.

PR #572, opened 2019-07-01
task/DES-818: Add search manager classes and consolidate file listing views.

jarosenb:
This PR sets the stage for future search enhancements by adding a set of search managers to set up queries and execute search calls. This eliminates redundant code in Site Search and the Data Depot. File listings have been updated to use a lookup manager (ported from CEP) so that searches and file listings can use the same view/API. 

Other changes:
- The `api/public_data` app has been combined with the `api/agave` app so that all listings use the same set of views, with a listing/search manager selected based on the route.
- `public_search_index.py`  has been broken up and the code has been moved to the more appropriate `lib/elasticsearch` module. The `Publication` and `LegacyPublication` classes are replaced by `BaseESPublication` and `BaseESPublicationLegacy` which inherit from `BaseESResource`, so there's now only one base implementation of an Elasticsearch wrapper class with custom getter/setter methods.
- Searching in the My Projects listing now filters the project list instead of searching for project files; we can revisit this when we have a better way of displaying metadata in project search results.


designsafe/apps/api/agave/filemanager/published_files.py
```diff
@@ -0,0 +1,52 @@
...
+from django.conf import settings
+from .base import BaseFileManager
+from designsafe.apps.api.agave.filemanager.agave import  AgaveFileManager
+from designsafe.apps.api.exceptions import ApiException
+
+
+logger = logging.getLogger(__name__)
+
+class PublishedFileManager(AgaveFileManager):
+    NAME = 'community'
```
Strmiska: Should this name be `'published'`?

designsafe/apps/api/agave/filemanager/publications.py
```diff
@@ -0,0 +1,107 @@
...
+            {'project._exact': {'order':'asc', 'unmapped_type': 'keyword'}},
+            {'created': {'order': 'desc', 'unmapped_type': 'long'}}
+        )
+        listing_search = listing_search.extra(from_=offset, size=limit)
+
+        res = listing_search.execute()
+        children = []
+        for hit in res:
+            try:
+                pid = hit.projectId
```
Strmiska: I don't think we are using `pid` in here.

jarosenb: I'm trying to distinguish between DES and NEES publications by checking for a `projectId` attribute- I'll refactor to make that more readable.

designsafe/apps/api/search/searchmanager/published_files.py
```diff
@@ -0,0 +1,82 @@
+"""
+.. module: portal.apps.search.api.managers.private_data_search
```
SilversunKSauri: I noticed that this manager plus a couple of the other managers (publications and shared data) have this "module" data wrong.  Looks like it was copied from the private_data manager and then not updated for which manager this is.  Not a big deal, since I don't know what uses this, but something I noticed.

jarosenb: Thanks, this should be fixed now.

PR #573, opened 2019-07-01
task/DES-1143 - Modify sitemap to use absolute path of each URL

fnets:
It can be a SEO problem to have the site map show pages without `www` but then have the server redirect those requests to `www` pages. Google is smart enough to see the redirect but the content can be considered duplicate or not authoritative. Updated the sitemap page URLs to include the `www` since that is how our server is configured to redirect all requests.

### DIFF
--- a/designsafe/sitemaps.py
+++ b/designsafe/sitemaps.py
@@ -53,12 +53,17 @@ from designsafe.apps.dashboard import urls as dashboard_urls
 from designsafe.apps.box_integration import urls as box_integration_urls
 from designsafe.apps.dropbox_integration import urls as dropbox_integration_urls
 from designsafe.apps.googledrive_integration import urls as googledrive_integration_urls
+from cms.sitemaps import CMSSitemap
 
 # Home
 class HomeSitemap(sitemaps.Sitemap):
     priority = 1.0
     changefreq = 'weekly'
 
+    def get_urls(self, site=None, **kwargs):
+        site = Site(domain='www.designsafe-ci.org')
+        return super(HomeSitemap, self).get_urls(site=site, **kwargs)
+    
     def items(self):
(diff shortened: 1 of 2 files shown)


designsafe/sitemaps.py
```diff
@@ -184,3 +201,14 @@ def items(self):
 
     def location(self, item):
         return item
+
+
+class CMSSitemap_modified(CMSSitemap):
```
rstijerina: How about
```suggestion
class DesignSafeCMSSitemap(CMSSitemap):
```
?

fnets: Sure! I'll need to change that name in a couple places.

fnets: done.

designsafe/sitemaps.py
```diff
@@ -184,3 +201,14 @@ def items(self):
...
         return item
+
+
+class DesignSafeCMSSitemap(CMSSitemap):
+    priority = .7
+    changefreq = 'weekly'
+
+    def get_urls(self, site=None, **kwargs):
+        site = Site(domain='www.designsafe-ci.org')
+        return super(CMSSitemap, self).get_urls(site=site, **kwargs)
```
jarosenb: didn't catch this earlier, but I think this should be `super(DesignSafeCMSSitemap, self)`

rstijerina: ah, yep good catch. @fnets can you take a look?

fnets: Yes, one sec.

fnets: Done. Had to open a separate PR, since this is already merged.  
https://github.com/DesignSafe-CI/portal/pull/581

PR #574, opened 2019-07-02
Bug/DES-1146 - Fix 403 response from published file listing

Strmiska:
removed conditional api params
`file.getMeta()` directs to a view that requires login. Replaced with `file.fetch()` which will use `PublicDataListView` instead.


jarosenb:
Works for me locally.

PR #576, opened 2019-07-02
Bug/DES-1064 - Email Confirmation View Rejection Message

fnets:
Users were getting a misleading error message when confirming the email.  Their emails were being confirmed in TAS, but an error message was telling them that it wasn't. 

This was because of an early `get_user_model` call that was causing an exception.  This call has been moved to a place where the exception will be handled better.  It will likely still fail at email confirmation, but will succeed later in the `agave_oauth_callback` function.

I also improved some messaging, so that logs will show when a user's email account has been confirmed in TAS successfully.


fnets:
Closing this pull request to address merge conflicts.

PR #585, opened 2019-07-18
bug/DES-1171: Update project model in elasticsearch

jarosenb:
We needed to add some fields to support newer style projects. After merging, we'll need to run `python manage.py swap_reindex --index='projects'`.

### DIFF
--- a/designsafe/apps/api/tasks.py
+++ b/designsafe/apps/api/tasks.py
@@ -701,7 +701,7 @@ def reindex_projects(self):
             in_loop = False
         else:
             for project in listing:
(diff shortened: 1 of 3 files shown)


designsafe/apps/api/tasks.py
```diff
@@ -698,8 +701,8 @@ def reindex_projects(self):
             in_loop = False
         else:
             for project in listing:
-                index_or_update_project.apply_async(args=[project.uuid], queue='api')
-
+                    index_or_update_project.apply_async(args=[project.uuid], queue='api')
```
Strmiska: Accidental tab? Otherwise LGTM!

PR #587, opened 2019-07-19
fix publication display in task/DES-1133

jarosenb:
The listing needed to be tweaked to use the new Published File manager.


designsafe/static/scripts/workspace/controllers/data-browser.js
```diff
@@ -1,4 +1,5 @@
 import _ from 'underscore';
+import { FieldReconPublishedViewComponent } from '../../data-depot/components/published/published-view.component';
```
rstijerina: Is this import necessary?

jarosenb: Weird, I have no idea how this line got here. I'll remove it.

PR #590, opened 2019-07-25
task/DES-1165 - Update Canonical URLs on Published Project Templates

fnets:
**Description:**  
Update Canonical URLs on Published Project Templates to an absolute URLs to match the sitemap. Also updated the sitemap for NEES Published Project URLs to have on a single slash. This is to improve discoverability by Google.

**Migrations:**  
No migrations.

**Expected behavior:**
1. Canonical URLs on Published projects should be absolute URLs that match the URL of the published project. If you access the project by navigating through the portal, there may be a double slash in the URL before the Project ID. You can just remove one of those slashes to reach the actual URL of the project.
2. Sitemap should have a single slash after all published projects liste

### DIFF
--- a/designsafe/sitemaps.py
+++ b/designsafe/sitemaps.py
@@ -180,6 +180,7 @@ class ProjectSitemap(sitemaps.Sitemap):
             
             for proj in projects['children']:
                 if 'project' in proj:
+                    # designsafe projects
                     subpath = {
                         'root' : reverse('designsafe_data:data_depot'),
                         'project' : proj['project'],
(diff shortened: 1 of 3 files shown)


designsafe/sitemaps.py
```diff
@@ -193,7 +194,7 @@ def items(self):
                         'project' : proj['path'],
                         'system' : proj['system']
                     }
-                    projPath.append('{root}public/{system}/{project}'.format(**subpath))
+                    projPath.append('{root}public/{system}{project}'.format(**subpath))
                 else:
```
Strmiska: I was a little confused with this part. If `project` in this case is the project id `project: "PRJ-XXXX"` why are we removing the `/` between `PRJ-XXXX` and `designsafe.storage.published`? I'm only asking because in the href of `base.html` we still have the slash after the storage system and before the projectId like so:
`https://www.designsafe-ci.org/data/browser/public/designsafe.storage.published/{{projectId}}`

fnets: This is for NEES projects, which includes a leading slash.  One of Four Kitchen's asks is to have only a single slash for these sitemap URLs.  This is so the canonical URLs match the sitemap URLs, which would match the project's URL.  So, to make it a single slash in the sitemap, I had to get rid of the slash here.

PR #624, opened 2019-09-10
task/DES-758: Force users to update profile.

jarosenb:
user_update_force_small

- Combine User Profile and Professional Profile into a single view.
- Add an `update_required` bool field to the `DesignSafeProfile` model defaulting to `True`.
- Add a middleware that redirects users to the profile update form if they try to access the dashboard, data depot, workspace, or recon portal. If the profile is successfully updated, the `update_required` flag is switched to `False`.


designsafe/middleware.py
```diff
@@ -33,6 +34,29 @@ def process_request(self, request):
...
 
+class DesignsafeProfileUpdateMiddleware:
+
+    """
+    Middleware to check if a user's profile has the update_required flag set to 
+    True, and force them to update their profile if so.
+    """
+
+    def process_request(self, request):
+        blocked_path = request.path.startswith('/data')\
```
xirdneh: Prefix can also be a tuple of prefixes to lookup.
```suggestion
        blocked_path = request.path.startswith(
            ("/data", "/applications", "/rw/workspace", "/recon-portal", "/dashboard")
        )
```

designsafe/middleware.py
```diff
@@ -33,6 +34,29 @@ def process_request(self, request):
...
+            or request.path.startswith('/applications')\
+            or request.path.startswith('/rw/workspace')\
+            or request.path.startswith('/recon-portal')\
+            or request.path.startswith('/dashboard')
+        if request.user.is_authenticated:
```
xirdneh: Consider falttening the nested ifs for less cyclomatic complexity:
```suggestion
        if request.user.is_authenticated and request.user.profile.update_required and blocked_path:
            # show warning
        elif request.user.is_authenticaated:
            return None
```

designsafe/apps/accounts/views.py
```diff
@@ -35,6 +35,11 @@ def manage_profile(request):
     """
     django_user = request.user
     user_profile = TASUser(username=request.user.username)
+    try:
+        ds_profile = DesignSafeProfile.objects.get(user__id=django_user.id)
+    except DesignSafeProfile.DoesNotExist:
+        logout(request)
```
xirdneh: I'm not understanding why we need to logout a user here. Can you explain?
Thanks.

jarosenb: This comes directly from the old `manage_pro_profile` profile view that is now combined with `manage_profile`. I'm not sure what the rationale is but I didn't want to change the logic.

PR #637, opened 2019-10-03
task/DES-1128: notifications preferences updates

SilversunKSauri:
Removed notification preferences tab from Manage Account, since all email notifications are handled through mailman now.

Added Manage Account to user menu dropdown at top right.


fnets:
LGTM

PR #642, opened 2019-10-11
DES-1297 - Add DOI Metatags for new Publication Templates

fnets:
Updated `get_context_data` to bring DOIs into metatags for new publications (`version: 2`).

### DIFF
--- a/designsafe/apps/data/views/base.py
+++ b/designsafe/apps/data/views/base.py
@@ -194,9 +194,9 @@ class DataDepotPublishedView(TemplateView):
         context['projectId'] = pub.projectId
         context['citation_title'] = pub.project.value.title
         context['citation_date'] = pub.created
-        if pub.project.value.to_dict().has_key('dois'):
+        if pub.project.value.to_dict().get('dois') != None: #This is for newer publications
             context['doi'] = pub.project.value.dois[0]
-        if pub.project.to_dict().has_key('doi'):
+        elif pub.project.to_dict().get('doi') != None: #This is for older publications
             context['doi'] = pub.project.doi
         context['keywords'] = pub.project.value.keywords.split(',')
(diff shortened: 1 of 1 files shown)


designsafe/apps/data/views/base.py
```diff
@@ -194,7 +194,10 @@ def get_context_data(self, **kwargs):
         context['projectId'] = pub.projectId
         context['citation_title'] = pub.project.value.title
         context['citation_date'] = pub.created
-        context['doi'] = pub.project.doi
+        if pub.project.value.to_dict().has_key('dois'):
+            context['doi'] = pub.project.value.dois[0]
+        if pub.project.to_dict().has_key('doi'):
+            context['doi'] = pub.project.doi
```
jarosenb: Because the Elasticsearch model has fields for both `project.value.dois` and `project.doi`, both of these if statements are going to execute every time. You'll want to check whether `doi` and `dois` are an empty string and an empty list, respectively. I don't know the pipeline well enough to know if there are cases where both of those fields are non-empty, and which should take precedence if so.

fnets: Updated to address this.

Strmiska: Within the publication object you can check for `pub.version` only the newer publications will have this version and it should be set to `2`. All other projects from before the Projects update will not have this field. To be clear older projects will have `pub.project.doi` newer projects will have `pub.project.value.dois`

fnets: I wanted to connect the conditionals to the actual fields that we are assigning, in case they change but we keep using `version: 2`.  The `get` should return `None` if the field that it's checking for is empty.

jarosenb: Looks good to me now.

PR #645, opened 2019-10-14
Task/DES-1258(1241) - Add FR filter to published listing and modify formatting

Strmiska:
- Make Field Research a filtered model type in the published listing
- Change the display format of Field Research and Hybrid Simulation in the published listing
- Remove bold from mission labels


jarosenb:
I'll roll the styling changes into the pub filter overhaul in DES-1291.

PR #648, opened 2019-10-16
Task/DES-1308 - Fix bugs with adding self to Team Members section

Strmiska:


designsafe/static/scripts/projects/components/edit-project/edit-project.component.js
```diff
@@ -376,11 +374,13 @@ class EditProjectCtrl {
             });
         }
         if (this.form.teamPrune) {
-            // this.form.teamPrune = projectData.teamMembers;
             this.form.teamPrune.forEach((ent) => {
                 projectData.teamMembers.push(ent.username);
             });
         }
+        if (!projectData.teamMembers.concat(projectData.coPis, [projectData.pi]).includes(this.form.creator.username)) {
```
rstijerina: Since "creator" in this component is always the current user, this makes it so that you can't ever remove yourself from a project even if you didn't create it. Is this intended?

Strmiska: This was intentional. We wanted to make it so users could not accidentally remove themselves from a project, but now that I'm thinking about it, we could flash a warning or something saying "Hey you're about to remove yourself from the project" before continuing.

designsafe/static/scripts/projects/components/edit-project/edit-project.component.js
```diff
@@ -425,23 +431,28 @@ class EditProjectCtrl {
...
-                this.$state.go(
-                    'projects.view.data',	
-                    {	
-                        projectId: project.uuid,	
-                        filePath: '/',	
-                        projectTitle: project.value.title	
-                    },	
-                    {reload: true}	
-                );
+        this.modalInstance.result.then((res) => {
```
rstijerina: This modalInstance response handler should be in the same if block as where it is instantiated, since if it is never instantiated this throws an error, i.e. when the user does not remove themselves.

The `this.savePrj` can stay down here as it was before.

Strmiska: Nah I’ll get it sorted out tomorrow morning. We can go to dev without it.

designsafe/static/scripts/projects/components/confirm-message/confirm-message.component.js
```diff
@@ -10,7 +10,7 @@ class ConfirmDeleteCtrl {
     }
 
     $onInit() {
-        this.ent = this.resolve.options.entity;
+        this.message = this.resolve.message;
     }
 
     delete () {
```
fnets: This function might also need to forward the user back to the project listing.  As it stands now, after I remove myself from a project I'm returned to the project.  And, I still have the ability to add myself back as a PI.

designsafe/static/scripts/projects/components/edit-project/edit-project.component.js
```diff
@@ -424,25 +421,48 @@ class EditProjectCtrl {
...
+                    message: () => "Are you sure you want to remove yourself from the project?",
+                },
+                size: 'sm'
+            });
+            this.modalInstance.result.then((res) => {
+                if (!res) {
+                    this.ui.busy = false;
+                } else {
+                    this.savePrj(projectData).then((project) => {
+                        if (this.project) {
```
fnets: This function might also need to forward the user back to the projects listing.  As it stands now, after I remove myself from a project I'm returned to the project.  And, I still have the ability to add myself back as a PI.  It's not until I go to the projects listing that I'm unable to access the project anymore.

Strmiska: Ah good point

fnets: It works!

designsafe/static/scripts/projects/components/manage-experiments/manage-experiments.component.js
```diff
@@ -297,11 +297,11 @@ class ManageExperimentsCtrl {
             this.editExpForm = {};
             this.ui.showEditExperimentForm = false;
         }
```
fnets: This might be outside the scope of this PR, but I'm noticing that when I add the first experiment or FR I need to exit the modal and re-enter it.  This doesn't happen when I'm adding my second experiment/FR.  Feel free to ignore.

Strmiska: Interesting, thanks for testing that. I'm going to swing by later so you can show me what it's doing. Might need to make a Jira for that today.

PR #652, opened 2019-10-18
Task/DES-1321 - Fixing bugs in collections modal for FR

Strmiska:
- Referenced Data, Instrument, and Observation Type have issues with displaying, adding, and removing fields correctly
- These fields will add blank inputs to metadata which become impossible to remove, and in some cases cause the collection to become stuck

We needed to make sure all invalid form inputs are cleared from the collection when saving and updating.

### DIFF
--- a/designsafe/static/scripts/projects/components/manage-field-recon/collections/manage-field-recon-collections.component.js
+++ b/designsafe/static/scripts/projects/components/manage-field-recon/collections/manage-field-recon-collections.component.js
@@ -47,8 +47,6 @@ class ManageFieldReconCollectionsCtrl {
 
         this.ui = {
             loading: false,
-            wantsToDelete: {},
-            confirmDelete: {},
         };
 
         this.data = {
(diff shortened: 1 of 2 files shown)


designsafe/static/scripts/projects/components/manage-field-recon/collections/manage-field-recon-collections.component.js
```diff
@@ -290,18 +288,36 @@ class ManageFieldReconCollectionsCtrl {
...
             longitude: this.form.longitude,
             latitude: this.form.latitude,
             elevation: this.form.elevation,
-            instruments: this.form.instruments,
-            referencedDatas: this.form.referencedDatas,
+            instruments: [],
+            referencedDatas: [],
             description: this.form.description,
         };
+        this.form.observationTypes.forEach((obs) => {
```
jarosenb: Instead of creating an empty array and then populating it, could you do something like `observationTypes = this.form.observationTypes.filter(...)`?

Strmiska: Ah yeah that would look cleaner.

PR #654, opened 2019-10-21
task/DES-1191: Add Project ID to Project Invitation E-mail

owaisj:
set_project_id is called right before email_collaborators . 
email_collaborator_added_to_project has project_id as parameter

### DIFF
--- a/designsafe/apps/api/projects/views.py
+++ b/designsafe/apps/api/projects/views.py
@@ -2,7 +2,7 @@
 import copy
 import logging
 import json
-from celery import group
+from celery import group, chain
 from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.http.response import HttpResponseForbidden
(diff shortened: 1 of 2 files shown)


designsafe/apps/api/projects/views.py
```diff
@@ -266,8 +266,10 @@ def post(self, request):
             prj.add_team_members([request.user.username])
 
         # Email collaborators
+        tasks.set_project_id.apply_async(args=[prj.uuid], queue="api")
         tasks.email_collaborator_added_to_project.apply_async(
             args=[
+                prj.project_id,
```
jarosenb: Does the `project_id` passed to `email_collaborator_added_to_project` depend on the previous call to `set_project_id`? If so, this looks like a potential race condition.

Strmiska: Oh yeah that would be an issue when creating a project for the first time. I'm guessing we could make these tasks into a chain like:
`(tasks.set_project_id.s(...)) | tasks.email_collaborator_added_to_project.s(...)).apply_async()`

jarosenb: Another complication: `set_project_id` won't update the `prj.id` value of a project that's already been retrieved from the database, so you'll have to make sure to call `refresh_from_db()` on it.

designsafe/apps/api/tasks.py
```diff
@@ -915,21 +915,23 @@ def set_facl_project(self, project_uuid, usernames):
         logger.debug('set facl project: {}'.format(res))
 
 @shared_task(bind=True, max_retries=3, default_retry_delay=60)
-def email_collaborator_added_to_project(self, project_title, project_url, team_members_to_add, co_pis_to_add):
-
+def email_collaborator_added_to_project(self, project_uuid, project_title, project_url, team_members_to_add, co_pis_to_add):
+    from designsafe.apps.projects.models.agave.experimental import ExperimentalProject
```
jarosenb: As a heads-up, we have a lot of different project types (experimental, simulation, field research, and growing). You won't need to worry about it here if you use Celery chaining, but if you need to in the future you'll want to do a lookup to make sure you're using the correct model. There's an example of how to do it in the POST method of `designsafe/apps/api/projects/views.py` where `prj` is defined.

PR #656, opened 2019-10-21
Task/DES-1311 - Change published and project listing header

Strmiska:
- Remove old published listing component/template
- Change header labels in project and published listings


Strmiska:
> Looks good. May have missed a ref to `ddPublicationListingComponent` in `publications.spec.js`

Ah thanks I did miss that! I think it's safe to get rid of that too.

PR #665, opened 2019-10-28
Task/DES-1265: Multiple metadata values need to align top with the key, not align center

owaisj:
Updated publication-preview styling

**Before**

**After**

_This image is at a different breakpoint_
Screen Shot 2019-10-29 at 2 07 23 PM


jarosenb:
Can you please add before/after pictures to the PR description?

PR #675, opened 2019-11-01
Task/DES-1347: Create Equipment List JSON

owaisj:
JSON file with equipment divided into arrays by type


owaisj:
He gave it the okay and outlined the next steps for image 7 on DES-1329

PR #677, opened 2019-11-04
Task/DES-1333: Create Author Information Modal

owaisj:
Replaced `ds-author-list` directive with series of anchor elements for contact information modals. Only uses object that has email and institution information.

Screen Shot 2019-11-04 at 3 07 17 PM


owaisj:
As per Craig's request, I added a ORCID ID field to users that have it added to their profile.
Screen Shot 2019-11-06 at 5 25 02 PM

Strmiska:
Ack, forgot to tell you that we need the PI and CoPI fields in the project header of the publication to have the author popup as well. Otherwise LGTM

PR #680, opened 2019-11-06
Task/DES-1290: Add DOI information as identifier meta tag

owaisj:
Added identifier as described in JIRA task as well as Dublin Core identifier tag.


jarosenb:
Will this need to handle publications that have `dois` instead of `doi`?

owaisj:
I added a conditional to the base.html template

PR #682, opened 2019-11-06
task/DES-1284 - Logging for Previews of files

fnets:
This will allow User Services to differentiate between a preview of a file and a download.  It also updates the logs with the IP of the request.


jarosenb:
I'm tagging this as WIP until we know we're 100% in the clear to be logging IPs like this.

fnets:
@jarosenb DES-1355 should address this.

fnets:
Removed the IP logging from this PR so that it can get out of WIP.  Depending on how the IP logging shakes out, we can add it back in later.

PR #684, opened 2019-11-07
Task/DES-1257: Author and Keyword Metatags on Publications

owaisj:
Tested using PRJ-2537, PRJ-2621, and PRJ-2550. 
Author names are encoded into UTF-8 to avoid Django unicode encode error in characters like ó seen in author list of 2550.

### DIFF
--- a/designsafe/apps/data/templates/data/data_depot.html
+++ b/designsafe/apps/data/templates/data/data_depot.html
@@ -3,6 +3,9 @@
 
 {% block google_citation_meta %}
     <!--data/data_depot.html-->
+    <meta name="keywords" content='{{ keywords|join:", " }}'> <!-- related experiements, subject; used in search -->
+    <meta name="authors" content="{% for author in authors %}{{author.full_name}}; {% endfor %}"> <!-- author names -->
+    
     <meta name="citation_title" content="{{citation_title}}">
     <meta name="citation_publication_date" content="{{citation_date}}">
     <meta name="citation_doi" content="{{doi}}">
@@ -12,8 +15,8 @@
     <meta name="citation_keywords" content="{{keyword}}">
     {% endfor %}
     {% for author in authors %}
-        <meta name="citation_author" content="{{author.full_name}}">
-        <meta name="citation_author_institution" content="{{author.institution}}">
+        {% if author.full_name %}<meta name="citation_author" content="{{author.full_name}}">{% endif %}
+        {% if author.institution %}<meta name="citation_author_institution" content="{{author.institution}}">{% endif %}
     {% endfor %}
     <meta name="citation_abstract_html_url" content="">
     <!--data/data_depot.html-->
--- a/designsafe/apps/data/templates/data/data_depot.html
+++ b/designsafe/apps/data/templates/data/data_depot.html
@@ -15,8 +15,8 @@
     <meta name="citation_keywords" content="{{keyword}}">
     {% endfor %}
     {% for author in authors %}
-        {% if author.full_name %}<meta name="citation_author" content="{{author.full_name}}">{% endif %}
-        {% if author.institution %}<meta name="citation_author_institution" content="{{author.institution}}">{% endif %}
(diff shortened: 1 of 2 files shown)


designsafe/apps/data/templates/data/data_depot.html
```diff
@@ -12,8 +15,8 @@
     <meta name="citation_keywords" content="{{keyword}}">
     {% endfor %}
     {% for author in authors %}
-        <meta name="citation_author" content="{{author.full_name}}">
-        <meta name="citation_author_institution" content="{{author.institution}}">
+        {% if author.full_name %}<meta name="citation_author" content="{{author.full_name}}">{% endif %}
```
fnets: For consistency's sake, can you remove these conditionals?  I'd rather know if these tags are empty.

designsafe/apps/data/views/base.py
```diff
@@ -199,12 +199,27 @@ def get_context_data(self, **kwargs):
         elif pub.project.to_dict().get('doi') != None: #This is for older publications
             context['doi'] = pub.project.doi
         context['keywords'] = pub.project.value.keywords.split(',')
-        context['authors'] = [{
-            'full_name': '{last_name}, {first_name}'.format(
-                last_name=user['last_name'], first_name=user['first_name']
-            ),
-            'institution': getattr(getattr(user, 'profile'), 'institution', '')
-        } for user in getattr(pub, 'users', [])]
+        if pub.__contains__('users'):
```
fnets: Do you mind writing these `__contains__` statements as `if users in pub`?  It's a little more pythonic and readable.

fnets: Also on line 209.

designsafe/apps/data/views/base.py
```diff
@@ -199,12 +199,27 @@ def get_context_data(self, **kwargs):
...
+            context['authors'] = [{
+                'full_name': '{last_name}, {first_name}'.format(
+                    last_name=user['last_name'].encode('utf-8'), first_name=user['first_name'].encode('utf-8')
+                ),
+                'institution': getattr(getattr(user, 'profile'), 'institution', '')
```
fnets: Does writing this as `user.profile.institution` work?

designsafe/apps/data/views/base.py
```diff
@@ -199,12 +199,27 @@ def get_context_data(self, **kwargs):
...
+                ),
+                'institution': getattr(author, 'inst', '')
+            } for author in getattr(pub, 'authors',[])]
+        else:
+            context['authors'] = [{
+                'full_name': '{last_name}, {first_name}'.format(
+                    last_name=author['lname'].encode('utf-8'), first_name=author['fname'].encode('utf-8')
+                ),
+                'institution': getattr(author, 'inst', '')
+            } for author in getattr(pub.project.value, 'teamOrder', [])]
```
fnets: Does `pub.project.value.teamOrder` work?

designsafe/apps/data/views/base.py
owaisj: I used __contains__ because when I try the `in` operator, I get an error from Django saying that the BaseESPublication model is not iterable. Would using the `to_dict()` method be okay?

designsafe/apps/data/views/base.py
owaisj: I used `getattr` because without it, if user.profile.institution is undefined, an AttributeError exception is thrown.

owaisj: I added the default parameter to this line in a new commit because without it, this function also throws an AttributeError.

designsafe/apps/data/views/base.py
Strmiska: `pub` on it's own is not a dictionary so it does not have the standard `get` methods until you do something like `pub.to_dict()`

designsafe/apps/data/views/base.py
fnets: Thanks, I think that should work!

PR #689, opened 2019-11-08
Task/DES-1364: Remove "Team Members" from Project Citations

owaisj:
Replaced `this.auths` with values from coPis and PI fields instead of authors

Example shown uses [PRJ-1434](https://www.designsafe-ci.org/data/browser/public/designsafe.storage.published/PRJ-1434)
Screen Shot 2019-11-08 at 10 41 22 AM


Strmiska:
LGTM

PR #690, opened 2019-11-08
Task/DES-1255: Metatags for multiple citations

owaisj:
Tested with
- PRJ-1434 (Experimental)
- PRJ-1648 (Experimental)
- PRJ-2319 (Simulation)
- PRJ-2548 (Simulation)
- PRJ-2397 (Field Recon)

### DIFF
--- a/designsafe/apps/data/views/base.py
+++ b/designsafe/apps/data/views/base.py
@@ -212,6 +212,7 @@ class DataDepotPublishedView(TemplateView):
         context['experiments'] = getattr(pub, 'experimentsList', [])
         context['missions'] = getattr(pub, 'missions', [])
         context['simulations'] = getattr(pub, 'simulations', [])
+        context['hybrid_simulations'] = getattr(pub, 'hybrid_simulations',[])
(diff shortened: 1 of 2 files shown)


designsafe/apps/data/views/base.py
```diff
@@ -207,7 +209,9 @@ def get_context_data(self, **kwargs):
         } for user in getattr(pub, 'users', [])]
         context['publication'] = pub
         context['description'] = pub.project.value.description
-        
+        context['experiments'] = getattr(pub, 'experimentsList', [])
+        context['missions'] = getattr(pub, 'missions', [])
+        context['simulations'] = getattr(pub, 'simulations', [])
         if self.request.user.is_authenticated:
```
Strmiska: I think you might need to add `hybridSimulations` in here as well. (we don't have any published at the moment, but the model exists in our system.)

owaisj: Added in a new commit. I used hybrid_simulation.py in apps/projects/models/agave as a reference.

PR #694, opened 2019-11-14
Task/DES-1374: Get all Django unit tests working

owaisj:
Edited methods and imports to reference changed or deleted modules. Some tests are skipped using unittest library.


rstijerina:
We probably shouldn't assert a test as True on a `try...except` failure. Do we know what kind of ConnectionError we're getting from Travis? Maybe we need to modify the test environment

PR #701, opened 2019-12-12
task/DES-935: fixing My Data file listing pagination

SilversunKSauri:
updated files-listing template for pagination and removing sorting by item type for now

This affects the My Data listing, files listing within a project, and potentially a file listing within a publication (though I have not tested that last one).

If you are on Keith's Walk Experiment Demo project, I copied my lotsoftestfiles directory into that project to test the pagination there, so that's available to anyone for testing as well.

### DIFF
--- a/designsafe/static/scripts/data-depot/components/data-depot-listing/files-listing.template.html
+++ b/designsafe/static/scripts/data-depot/components/data-depot-listing/files-listing.template.html
@@ -1,7 +1,10 @@
 <div class="table-responsive" data-ng-class="{'preview-wrapper': $ctrl.browser.showPreviewListing}">
-    <div data-ds-infinite-scroll
-         data-scroll-bottom="$ctrl.scrollToBottom()"
-         data-bottom-height="0"
+    <div
+        class="ds-table-display-wrapper"
+        data-ds-infinite-scroll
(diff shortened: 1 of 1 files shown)


rstijerina:
Goal of this PR is not pagination, but confining results to a single fixed-height scroll window correct?

SilversunKSauri:
Apparently adding the class to confine the results to the fixed-height box was enough to fix the pagination.  So technically the goal was to fix pagination, but by proxy the result box got created.  I think there was some issue with the triggering of the scrollToBottom function when the result box wasn't confined, but that's just my best guess at this point.  And the sorting is getting removed for now since it's creating a bad UX with the pagination.

rstijerina:
Hmm do you mean a paginated front end view? It's doing an infinite scroll for me

SilversunKSauri:
Ah, I get what you mean now.  Yes, it is infinite scroll, rather than a page-by-page version.  Before this it would only show the first 100 items, and now it will do the infinite scroll style pagination.  I don't know that we can get what we'd want from Agave to do a page-by-page pagination, but that's certainly one thought I had for trying to improve the UX and bring back the item type sorting in the future.

rstijerina:
Ah, gotcha. Cool, then yeah it works for me as intended

PR #705, opened 2020-01-07
Task/DES-1412 - Make users submit a ticket to republish or overwrite their publications

Strmiska:
This will be modified in the future, but for now we cannot allow users to intentionally or unintentionally overwrite their publications.

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-agreement.html
+++ b/designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-agreement.html
@@ -51,7 +51,8 @@
             This project has already been published.
             If you would like to update your existing publication or
             if you are having trouble viewing your publication please
-            <a href="/help/new-ticket/" target="_blank">submit a ticket</a>.
+            <a href="/help/new-ticket/?category=DATA_CURATION_PUBLICATION&subject=Requesting+update+to+existing+publication:+{{$ctrl.project.value.projectId}}"
+               target="_blank">submit a ticket</a>.
         </div>
         <div style="text-align: center">
                 <button class="btn btn-add"
--- a/designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-agreement.html
+++ b/designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-agreement.html
@@ -19,7 +19,12 @@
         publication that cannot be changed. If I have revisions this will result in a new
         version of the publication including a new DOI.
     </p>
-    <div ng-if="!$ctrl.published && !$ctrl.existingPub">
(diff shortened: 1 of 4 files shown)


SilversunKSauri:
Do you have an existing project/publication we can test this on?  The code looks good to me, just wasn't sure if there was a way you would like us to test otherwise.

designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-agreement.html
```diff
@@ -46,4 +46,20 @@ <h4 class="modal-title" style="border:none;">
                 </button>
         </div>
     </div>
+    <div data-ng-if="$ctrl.existingPub">
+        <div class="alert alert-warning">
+            This project has already been published.
+            If you would like to update your existing publication or
+            if you are having trouble viewing your publication please
+            <a href="/help/new-ticket/" target="_blank">submit a ticket</a>.
```
rstijerina: consider adding some prefilled ticket info here
```suggestion
            <a href="/help/new-ticket/?category=DATA_CURATION_PUBLICATION&subject=Update+existing+publication" target="_blank">submit a ticket</a>.
```

Strmiska: Good Idea! I'll add that in there.

Strmiska:
> Do you have an existing project/publication we can test this on? The code looks good to me, just wasn't sure if there was a way you would like us to test otherwise.

Totally missed this until recently! PRJ-2224 "Walk Experiment Demo" should be published in the local index. If you go through that publication to the very end after license the popup should display the modal.

designsafe/static/scripts/data-depot/components/projects/pipeline-publish/pipeline-publish.component.js
```diff
@@ -84,6 +86,12 @@ class PipelinePublishCtrl {
             });
         }
         this.publication = publication;
+        this.PublishedService.getPublished(this.project.value.projectId).then((resp) => {
+            let data = resp.data;
+            if (data.project) {
+                this.existingPub = true;
```
rstijerina: just noticed that this is a potential race condition since `this.PublishedService.getPublished` is dispatched as an asynchronous action. Perhaps set `this.existingPub = null` at the top of the `onInit` instead of `false`, and flag it `true` or `false` here. Then in the template, add a loading icon, or just disable the submit button until `this.existingPub !== null`

Strmiska: I see what you mean, I think that makes a little more sense. I'll get those changes added.

PR #712, opened 2020-01-23
Bug/DES-1404 - Remove field from experimental-data.json

Strmiska:
This is a hack to remove two dropdown options defined in a .json form. We can't delete the field from the .json because the existing published Experiments which selected this option will break. This .json file is referenced by both published and non published projects... Fixing this correctly will require updating all previously published experiments, as well as updating the way these fields are saved (so we don't reference a .json for both published and non published projects). At the time I've submitted this, I do not have access to modify published metadata locally, and there is an important user waiting for this change. I'll be making a ticket to fix the underlying issue.

### DIFF
--- a/designsafe/static/scripts/projects/components/manage-experiments/manage-experiments.component.js
+++ b/designsafe/static/scripts/projects/components/manage-experiments/manage-experiments.component.js
@@ -16,10 +16,15 @@ class ManageExperimentsCtrl {
     $onInit() {
         this.project = this.resolve.project;
         this.edit = this.resolve.edit;
-        
+
+        let remData = (data) => {
+            data.atlss = data.atlss.filter(a => a.name != "hybrid_simulation"); 
+            return data;
+        };
+
         this.efs = experimentalData.experimentalFacility;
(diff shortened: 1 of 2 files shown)


designsafe/static/scripts/projects/components/manage-experiments/manage-experiments.component.js
```diff
@@ -16,10 +16,15 @@ class ManageExperimentsCtrl {
     $onInit() {
         this.project = this.resolve.project;
         this.edit = this.resolve.edit;
-        
+
+        let remData = (data) => {
+            data.atlss = data.atlss.filter(a => a.name != "hybrid_simulation");
```
rstijerina: sanity check - will `data.atlss` always exist and be an array?

Strmiska: Yep, that "atlss" field is in a JSON object that was setup as a set of input field mappings for a form. This is the JSON that it's referencing.
https://github.com/DesignSafe-CI/portal/blob/master/designsafe/static/scripts/projects/components/manage-experiments/experimental-data.json

PR #717, opened 2020-02-04
Task/DES-1466 - Allow anonymous users to download and preview public files

Strmiska:
CSRF Token was being enforced for anonymous users viewing public file listings.

### DIFF
--- a/designsafe/apps/api/agave/views.py
+++ b/designsafe/apps/api/agave/views.py
@@ -15,7 +15,6 @@ from django.http import JsonResponse
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
 from django.utils.decorators import method_decorator
-from django.views.decorators.csrf import csrf_exempt
 from designsafe.apps.api.agave.filemanager.agave import AgaveFileManager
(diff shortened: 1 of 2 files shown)


designsafe/apps/api/agave/views.py
```diff
@@ -89,7 +90,7 @@ def get(self, request, file_mgr_name, system_id=None, file_path=''):
                                 encoder=AgaveJSONEncoder,
                                 safe=False)
 
-
+@method_decorator(csrf_exempt, name='dispatch')
```
rstijerina: this is exempting all of these methods as csrf_exempt -- you should apply it to only the public media views

Strmiska: Good catch, thanks! I missed that in the docs. It says I need to put the csrf_exepmtion on the dispatch, but later it says you also need to put csrf_protect/exempt on all the protected/non protected methods like PUT POST etc.. I'll get this sorted out.

Strmiska: Note: It may not work to exempt the CSRF token for downloads and previews, because this request uses a PUT method. Still working on this one...

PR #736, opened 2020-03-30
Task/DES-1552 - Download Published File Archives

Strmiska:
Created a "Publication Download" modal and updated a few of the templates. This should download any archive created for a publication or at least show an error message for the user to submit a ticket.

This can be tested locally if you have some published projects available in your index. Just make sure that the publication has an existing archive saved on corral. The path should look something like:
`/corral-repl/projects/NHERI/published/archives/PRJ-XXXX_archive.zip`

### DIFF
--- a/designsafe/static/scripts/projects/components/publication-download/publication-download.component.js
+++ b/designsafe/static/scripts/projects/components/publication-download/publication-download.component.js
@@ -11,6 +11,7 @@ class PublicationDownloadModalCtrl {
     $onInit() {
         this.loaded = null;
         this.error = null;
+        this.retrievingPostit = null;
         this.publication = this.resolve.publication;
         this.mediaUrl = this.resolve.mediaUrl;
         this.prjId = this.publication.project.value.projectId;
@@ -33,6 +34,7 @@ class PublicationDownloadModalCtrl {
     }
 
     download() {
+        this.retrievingPostit = true;
         this.archiveListing.download().then(function (resp) {
             let postit = resp.href;
             let link = document.createElement('a');
@@ -42,6 +44,9 @@ class PublicationDownloadModalCtrl {
             document.body.appendChild(link);
             link.click();
             document.body.removeChild(link);
+        }).finally(() => {
+            this.retrievingPostit = false;
+            this.cancel();
         });
     }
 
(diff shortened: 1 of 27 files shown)


rstijerina:
Is there a good example of a project that has this button? What would the url be for local dev?

Strmiska:
> @Strmiska Is there a good example of a project that has this button? What would the url be for local dev?

Sorry I should have posted that along with an image of where this button is! If you go into any of the published projects you will see a "Download Dataset" button at the top right next to the title. If you open the modal inside a published project it should show an error if the archive is missing (this should happen for a few publications locally). I think the first few experiments in the published listing should give a download option.

rstijerina:
>  I think the first few experiments in the published listing should give a download option.

Can you link me? Not managing to find one, but maybe it's my local setup?

Strmiska:
> > I think the first few experiments in the published listing should give a download option.
> 
> Can you link me? Not managing to find one, but maybe it's my local setup?

This should work, but if you can't preview the published files locally then you may not be able to see the archives as well.
https://designsafe.dev/data/browser/public/designsafe.storage.published//PRJ-2537

rstijerina:
> This should work, but if you can't preview the published files locally then you may not be able to see the archives as well.
> https://designsafe.dev/data/browser/public/designsafe.storage.published//PRJ-2537

Got it, thanks! Had to rebuild

designsafe/static/scripts/projects/components/publication-download/publication-download.component.js
```diff
@@ -0,0 +1,64 @@
...
+            this.arcData = resp;
+        }).finally(() => {
+            if (!this.arcData) {
+                this.error = true;
+            }
+            this.loaded = true;
+        });
+    }
+
+    download() {
```
rstijerina: can you add a spinner to the download button and disable the button while this function is running?

something like on `download`
`this.initiatingDownload = true;`

and in the template

```<button class="btn btn-add" ng-click="$ctrl.download()" ng-disabled="!$ctrl.loaded">
            <div class="btn-container">
                <i class="fa fa-spinner fa-spin" />
                <span class="curation-download"></span>
                &ensp;
                <span><strong>Download</strong></span>
            </div>
        </button>```

Strmiska: Sure, do you mean to show it spinning while the file is downloading or just while retrieving the agave postit? I think I we close the modal when the file starts to download don't you think. (these details weren't provided in the mockup)

rstijerina: Just while retrieving the postit. I think the rest of the behavior makes sense - closing the modal after the download starts

PR #737, opened 2020-03-31
Task/DES-1558: Add all "DOI(s) in Dataset" to project level information

owaisj:
Users can click on DOIs and be brought to projects using `$anchorScroll`. This also opens up project information tabs.

Additionally, a method is added as a check to see if there are related works available.


fnets:
lgtm, and works locally again.

PR #738, opened 2020-04-01
Bug/DES-1582 - Filter conditionally required field from error message

Strmiska:
A conditionally required entity type (social science or geoscience) was being returned incorrectly in an error message sent back to users even if they had the minimum requirements.

If a mission is published it must have:
- Planning Set
- Social Science OR Geoscience Set
- Files within all Sets


rstijerina:
Is there an underlying issue that is preventing the `data.missing` field from returning the correct data?

Strmiska:
> Is there an underlying issue that is preventing the `data.missing` field from returning the correct data?

No, other than that the Field research model has more complicated requirements than other publications.
This requirement checklist was meant to be a pretty straight forward method to check that there were "required" entities selected in a project and that they contained files. For every other project we just need to make sure that  x, y, and z are all present. With field recon this got kinda complicated by required that users should have a report OR a mission... Then if a mission(s) is selected each mission must have a planning entity and at least one social or geoscience entity. So I decided to use the same "checkRequirements" function to return the missing requirements. And then I would tweak the results/requirements specifically for the field_recon project type.

PR #740, opened 2020-04-06
Task/DES-1408: Dublin Core meta tags for Google Scholar citation

owaisj:
Added block of meta tags using Dublin Core schema

### DIFF
--- a/designsafe/apps/data/templates/data/data_depot.html
+++ b/designsafe/apps/data/templates/data/data_depot.html
@@ -68,6 +68,14 @@
     {% else %}
     <meta name="identifier" content="{{doi}}">
     {% endif %}
+    <!-- Dublin Core Metatags -->
+    <meta name="DC.identifier" content="{{doi}}">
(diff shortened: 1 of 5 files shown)


designsafe/apps/data/templates/data/data_depot.html
```diff
@@ -68,6 +68,14 @@
     {% else %}
     <meta name="identifier" content="{{doi}}">
     {% endif %}
+    <!-- Dublin Core Metatags -->
+    <meta name="DC.identifier" content="{{doi}}">
```
Strmiska, line 72: I'm not sure of the use case for this or the requirements, however there are two things I noticed.
- Is this identifier supposed to return something else if there is no DOI at the project level? In some cases there will be no DOI for the project; it will have multiple DOIs on the entities published within it.
- The DC meta content variables do not seem track changes in the view from one publication to another, but the meta content from the tags with names like `citation_XXX` do track.

PR #742, opened 2020-04-09
Fix various eslint-related warnings and errors

clawler:
I noticed that your travis-ci invocations of `rpm run eslint` were failing so this helps to fix some of those linter issues.


clawler:
lmk if you'd like a tinier more incremental set of changes - that said, these changes should be safe since eslint did them for me

PR #743, opened 2020-04-09
Make travis flake8 task work again

clawler:
Seems like Travis is still stuck on running unit tests on Python v2.7.

This attempts to upgrade Python to v3.6.

Will validate that this change got picked up in the Travis-CI builds.


.travis.yml
```diff
@@ -18,91 +18,22 @@ stages:
 - build
 matrix:
   include:
```
clawler, line 20: i organized these task configs so that they are in the same order as they are displayed in travis ci

.travis.yml
```diff
@@ -18,91 +18,22 @@ stages:
...
-    - '2.7'
+    - '3.6'
     stage: lint
     install:
-    - pip install -r requirements.txt
-    - find designsafe/settings/* -name *.sample.* | while read line; do cp $line ${line/sample\./};
-      done;
-    - "./manage.py collectstatic --settings=designsafe.settings.test_settings --no-input
-      --link -i demo > /dev/null"
+    - pip3 install -r requirements.txt
```
clawler, line 27: this makes it so that pip3 is explicitly used to install requirements like it is when building docker images for prod

PR #748, opened 2020-04-27
Task/DES-1616: Expose authors of Other/FR publications in search.

jarosenb:


jarosenb:
> Everything looks good, and the Data Depot search results return correctly. But I'm not getting hits on sitewide search.

The actual problem here is that pagination isn't working correctly in Site Search. Looking into it.

fnets:
LGTM!  Great job!

PR #750, opened 2020-05-01
hotfix: remove message attr ref from exception

rstijerina:
Exception.message was removed in python 3

### DIFF
--- a/designsafe/apps/applications/views.py
+++ b/designsafe/apps/applications/views.py
@@ -398,8 +398,8 @@ def call_api(request, service):
                 return HttpResponse('Unexpected service: %s' % service, status=400)
         except AgaveException as e:
             logger.error('Failed to execute {0} API call due to AgaveException={1}'.format(
-                service, e.message))
-            return HttpResponse(json.dumps(e.message), content_type='application/json',
+                service, e))
+            return HttpResponse(json.dumps(e), content_type='application/json',
                                 status=400)
         except HTTPError as e:
             try:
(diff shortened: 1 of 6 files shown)


designsafe/apps/applications/views.py
```diff
@@ -398,8 +398,8 @@ def call_api(request, service):
                 return HttpResponse('Unexpected service: %s' % service, status=400)
         except AgaveException as e:
             logger.error('Failed to execute {0} API call due to AgaveException={1}'.format(
-                service, e.message))
-            return HttpResponse(json.dumps(e.message), content_type='application/json',
+                service, e))
+            return HttpResponse(json.dumps(e), content_type='application/json',
```
nathanfranklin, line 402: Just testing a bit and see that `'AgaveException' is not JSON serializable`.  Looks like it should be `HttpResponse(e, ...)` for here and on Line 419.

designsafe/apps/notifications/views.py
```diff
@@ -96,7 +96,7 @@ def generic_webhook_handler(request):
     except (HTTPError, AgaveException, ApiException) as e:
         LOGGER.exception(
             "Could not find valid corresponding Agave job for interactive event")
-        return HttpResponse(json.dumps(e.message), content_type='application/json', status=400)
+        return HttpResponse(json.dumps(e), content_type='application/json', status=400)
```
nathanfranklin, line 99: same here and on line 117. an unrelated bug to what you were originally fixing in this pr.

rstijerina: good catch. will fix in a new PR.

PR #759, opened 2020-05-20
Task/DES-1540 & Bug/DES-1597: Author Modal Fixes

owaisj:
This PR fixes two bugs.
1) DES-1540: The `author` data is passed in to the modal component and then the orcid id is fetched from the back-end. Prior to this, the modal was only opened once the `UserService` promise was resolved.
2) DES-1597: Since the modal is triggered by an `ng-click` directive with '#' inside the href tag, the page was scrolling to the top. `href="javascript;;"` keeps the tag valid and stops the issue

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.js
+++ b/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.js
@@ -239,17 +239,12 @@ class PublicationPreviewFieldReconCtrl {
     }
 
     showAuthor(author) {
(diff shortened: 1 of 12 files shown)


designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.js
```diff
@@ -239,17 +239,12 @@ class PublicationPreviewFieldReconCtrl {
     }
 
     showAuthor(author) {
```
rstijerina, line 241: We should consider creating a higher order class for common publication action components, or at least a utility component to handle these shared functions across components.

PR #760, opened 2020-05-27
Bug/DES-1551: Delete button on filter in recon portal does not appear to work

owaisj:
This PR adds a method to clear filter text on click of the trash icon. It can be tested locally [here](https://designsafe.dev/recon-portal/).

### DIFF
--- a/designsafe/static/scripts/rapid/html/index.html
+++ b/designsafe/static/scripts/rapid/html/index.html
@@ -31,7 +31,7 @@
           <div class="input-group">
             <input type="text" class="form-control" ng-model="vm.filter_options.search_text" ng-change="vm.search()"></input>
             <span class="input-group-btn">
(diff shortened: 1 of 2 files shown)


designsafe/static/scripts/rapid/html/index.html
```diff
@@ -31,7 +31,7 @@
           <div class="input-group">
             <input type="text" class="form-control" ng-model="vm.filter_options.search_text" ng-change="vm.search()"></input>
             <span class="input-group-btn">
-              <button class="btn btn-default"> <i class="fa fa-trash"></i></button>
+              <button class="btn btn-default"> <i class="fa fa-trash" ng-click="vm.clear_filter_text()"></i></button>
```
nathanfranklin, line 34: what about just using `clear_filters()`.  This is totally debatable, but I feel like clearing all search options might be better

PR #762, opened 2020-06-04
task/DES-1650: Add infinite scroll to Data Depot move/copy modals.

jarosenb:
These modals never had infinite scroll to begin with, so this PR implements them. 

Test by setting `this.offset` or `$scope.offset` to a small number.

### DIFF
--- a/designsafe/static/scripts/ng-designsafe/components/modals/data-browser-service-move/data-browser-service-move.component.js
+++ b/designsafe/static/scripts/ng-designsafe/components/modals/data-browser-service-move/data-browser-service-move.component.js
@@ -20,7 +20,7 @@ class DataBrowserServiceMoveCtrl {
             files: this.files,
             names: {},
         };
-
+        this.currentParams = null;
         this.listing = this.initialDestination;
 
         this.state = {
@@ -62,6 +62,7 @@ class DataBrowserServiceMoveCtrl {
         this.$scope.$watch('currentOption', (opt) => {
             let conf = opt.conf;
             let params = opt.apiParams;
+            this.currentParams = params;
             this.state.reachedEnd = false;
             this.state.busy = true;
             this.offset = 0;
@@ -102,10 +103,11 @@ class DataBrowserServiceMoveCtrl {
 
     }
 
-    scrollToBottom() {
+    scrollToBottom(fileListing) {
         this.offset += this.limit;
         this.state.loadingMore = true;
         var cOption = this.$scope.currentOption;
+        cOption.conf.path = (fileListing || {}).path
         if (cOption.conf.system != 'projects'){
           this.state.listingProjects = false;
           this.FileListing.get(cOption.conf, cOption.apiParams, {offset: this.offset, limit: this.limit})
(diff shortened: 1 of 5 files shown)


designsafe/static/scripts/ng-designsafe/components/modals/data-browser-service-move/data-browser-service-move.component.js
```diff
@@ -26,9 +26,13 @@ class DataBrowserServiceMoveCtrl {
         this.state = {
             busy: false,
             error: null,
-            listingProjects: false
+            listingProjects: false,
+            listingProjects: false,
```
SilversunKSauri, line 30: Is there a reason `listingProjects: false,` is here twice?

jarosenb: Nope, I'll fix that.

SilversunKSauri:
I was just testing this locally and I'm getting some odd behavior.

Process I'm following:
1) Select a file in My Data and click "copy"
2) In the copy modal, I'm switching the area to My Data and then clicking on my "lotsotestfiles" directory, since that has almost 200 files/folders in it
3) I scroll down to see if it loads more, and I get one loading for another set of listing, but that's all it does, so it's still not giving me all the files I expect.  This folder has two subfolders alphabetically first and two alphabetically last, so that's my primary test is to see what happens with those four folders.  I could not scroll down to those last two folders.

I wanted to confirm the above behavior, and so I exited the copy modal and went into my "lotsotestfiles" directory to confirm the files I was expecting, then back to base My Data and did the same step 1 and 2 from above.  This time, I got the loading icon for another set, but when I scrolled down, I saw files in the modal from my base My Data which should not exist if I'm viewing my "lotsotestfiles" directory.  I'm not entirely sure what's going on or if it's something going wrong locally on my end, so I'm going to do some more testing to try to figure it out, but wanted to report it now while I was thinking about it.

designsafe/static/scripts/ng-designsafe/components/modals/data-browser-service-move/data-browser-service-move.component.js
```diff
@@ -26,9 +26,13 @@ class DataBrowserServiceMoveCtrl {
...
+            loadingMore: false,
+            reachedEnd: false,
         };
-        
+        this.offset = 0;
+        this.limit=100;
```
rstijerina, line 35: any reason the limit is `100` here and `20` in the copy modal?

designsafe/static/scripts/ng-designsafe/components/modals/data-browser-service-move/data-browser-service-move.component.js
rstijerina:
```suggestion
```

jarosenb:
> I was just testing this locally and I'm getting some odd behavior.


This should be fixed now, the `onBrowse` method wasn't getting some of the new params.

SilversunKSauri:
> This should be fixed now, the `onBrowse` method wasn't getting some of the new params.

Now it doesn't appear to be adding new files to the end of the listing at all.  It just seems to be showing me the first 100 without any new reloads.  Same steps as before, so I'm in a subdir of My Data that doesn't appear to be scrolling.  If it helps, I can either share my "large" folder with you on some project or in your My Data so you can test on it, or we can schedule a time to video chat and I can share screen.

PR #768, opened 2020-06-16
removing call to custom script from rapid initial migration

SilversunKSauri:
Rapid initial migration failing locally, and is unneeded, so removing the call to custom script form the migration.

### DIFF
--- a/designsafe/apps/rapid/migrations/0001_initial.py
+++ b/designsafe/apps/rapid/migrations/0001_initial.py
@@ -32,5 +32,4 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(add_rapid_event_types)
     ]


rstijerina:
you can get past this locally with `./manage.py migrate rapid --fake`

SilversunKSauri:
Yeah, that's what I've been doing, but Jake suggested we remove the custom script so we don't have to worry about it in the future.  I'm willing to go either way, honestly.

rstijerina:
I think I'd prefer to preserve the migration as is for provenance.

jarosenb:
> I think I'd prefer to preserve the migration as is for provenance.

How about removing the Elasticsearch call from the `add_rapid_event_types` function as a compromise? Or at least commenting it out. The way we set up indexes has fundamentally changed since that function/migration were written.

rstijerina:
> How about removing the Elasticsearch call from the `add_rapid_event_types` function as a compromise? Or at least commenting it out. The way we set up indexes has fundamentally changed since that function/migration were written.

Ah I see. If this step no longer has any bearing on `rapid` setup, and just commenting this out does not affect downstream migrations, then I'm ok with this change.

PR #773, opened 2020-07-07
Bug/DES-1662 - Fix issue with project file listings overloading agave with requests

Strmiska:
### Summary of Changes:
- **project-view.component.js** - Remove browser data from state transition
- **curation-directory.component.js** - Remove browser data from state transition...
- **publication-preview.component.js** - Clean up AbstractedListing function which builds the abstracted file listing for publication previews. No longer requires multiple requests for each file. Instead requests for necessary file information will be gathered on user interaction. (see file-categories.component.js)
- **publication-preview.template.html** - Merge` ng-if` and `ng-hide` directives
- **publication-preview-sim.component.js** - Clean up AbstractedListing function...
- **publication-preview-sim.template.html** - Merge `ng-if` and `ng-hide` directives...
- **publication-preview-hyb-sim.component.js** - Clean up AbstractedListing function...
- **publication-preview-hyb-sim.template.html** - Merge `ng-if` and `ng-hide` directives...
- **publication-preview-field-recon.component.js** - Clean up AbstractedListing function...
- **publication-preview-field-recon.template.html** - Merge `ng-if` and `ng-hide` directives…
- **pipeline-selection.component.js** - Clean up AbstractedListing function…
- **manage-categories.component.js** - Clean up AbstractedListing function...
- **manage-categories.template.html** - Do not show incomplete file listings in categories
- **file-listing** - Fixed issue where directory paths with periods `.` in the name would not render categories/tags for subsequent file listings. This was a result of an unnecessary check for a file extension on a parent directory.
- **file-category-selector.template.html** - Show loading spinner while busy
- **file-categories.component.js** - Refactor how we gather file information to manipulate and display tags on files. We will no longer gather the uuid of every file in the listing. Instead we will get the uuid when a user needs to change something on selected files. We will also render file tags based on the matching file path within the tag (instead of using the uuid). This will allow us to render tags without having to send requests for every single file (to get its uuid) in a project’s listing. Removed underscore `_.foo()` functions. Removed unnecessary `$scope.$apply()` functions.
- **file-categories.template.html** - Show loading spinner while busy

### Testing Changes:
These changes impact the working area for all project types (other, experimental, simulation, hybrid-sim, field-research), as well as published projects. In this PR you should notice improvements in loading times for the `Working Directory`, `Curation Directory`, `Publication Preview`, `Selection`(first step of publish pipeline), and the `Add Categories` modal. Any interactions with the file listings for projects should remain the same. This was pushed as a "Bug" because in larger projects/publications (like the one below) would often time out or have issues loading all of their files.

You should notice a _significant_ difference in load time between this PR and PROD for loading large projects/publications like this one:
https://designsafe.dev/data/browser/public/designsafe.storage.published//PRJ-2363
vs
https://www.designsafe-ci.org/data/browser/public/designsafe.storage.published//PRJ-2363

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-hyb-sim.component.js
+++ b/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-hyb-sim.component.js
@@ -20,7 +20,6 @@ class PublicationPreviewHybSimCtrl {
         this.readOnly = this.$state.current.name.indexOf('publishedData') === 0;
         this.projectId = this.ProjectService.resolveParams.projectId;
         this.filePath = this.ProjectService.resolveParams.filePath;
-        this.data = this.ProjectService.resolveParams.data;
         this.ui = {
             fileNav: true,
             loading: true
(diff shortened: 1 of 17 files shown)


jarosenb:
Once this is merged, is there any case where the `file-categories` component will render anything when you're in the Working Directory section of a project?

designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-hyb-sim.component.js
```diff
@@ -73,79 +70,52 @@ class PublicationPreviewHybSimCtrl {
                 });
                 child.setEntities(this.projectId, entities);
```
jarosenb, line 71: Does this `setEntities` call still do anything?

Strmiska:
> Once this is merged, is there any case where the `file-categories` component will render anything when you're in the Working Directory section of a project?

Yes, anywhere file tags/categories are displayed in a file listing we use the `file-categories` component. There is also the `file-category-selector` component which sounds redundant to have, but I think there is some functionality in that component that we did not want to have anywhere except the Curation Directory.

designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-hyb-sim.component.js
Strmiska: Yes, we add all associated entities to the child(file) of the listing in order to display the specific category and tags per category of each child(file).

jarosenb: In what situation would `browser.listing.children` be used in the publication preview? Naively it looks like the listings are all handled through `populateListing`.

Strmiska: When a user within the preview or published view of a project navigates into a directory we switch from the abstracted file listing `browser.listings` to the standard listing `browser.listing`. The children within these file paths can have tagged files as well.

jarosenb:
> > Once this is merged, is there any case where the `file-categories` component will render anything when you're in the Working Directory section of a project?
> 
> Yes, anywhere file tags/categories are displayed in a file listing we use the `file-categories` component. There is also the `file-category-selector` component which sounds redundant to have, but I think there is some functionality in that component that we did not want to have anywhere except the Curation Directory.

Got it, sorry, I was looking at a demo project that I thought had file tags but actually didn't.

designsafe/static/scripts/ng-designsafe/services/file-listing.js
```diff
@@ -65,7 +73,7 @@ export function FileListing($http, $q) {
             var myAsoc = _.find(self._entities[0]._links.associationIds,
```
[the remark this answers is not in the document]
Strmiska: Oh yeah, I saw this section in the file listing and it didn't make much sense to me. It looks like a very old way of managing file tags for the projects published before the revised projects/publications, and (from what I remember) file tags have never worked for those projects. I tried to remove this initially, but it started to cause some issues and I forgot about it. This should probably be removed, but I'll have to check that nothing changes in the older project's file listings as a result.

PR #776, opened 2020-07-22
Fix Simulations Not Rendering

owaisj:
This can be tested using [this publication](https://designsafe.dev/data/browser/public/designsafe.storage.published//PRJ-2290) locally. The `updateHeaderMetadata` method now adds reports to the header if they are published. 

In prod, as seen [here](https://designsafe-ci.org/data/browser/public/designsafe.storage.published//PRJ-2290), the publication gets stuck on loading.

### DIFF
--- a/designsafe/static/scripts/data-depot/services/published.service.js
+++ b/designsafe/static/scripts/data-depot/services/published.service.js
@@ -149,18 +149,20 @@ export class PublishedService {
         this.$window.document.getElementsByName('DC.creator')[0].content = authors;
 
         const entities = [];
+        let isSimulation = false;
         if (has(resp.data, 'experimentsList')) {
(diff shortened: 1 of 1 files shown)


designsafe/static/scripts/data-depot/services/published.service.js
```diff
@@ -160,7 +160,8 @@ export class PublishedService {
         }
 
         // Check for reports
-        if(has(resp.data, 'reports')) entities.push(...resp.data.reports);
+        const isSimulation = has(resp.data, 'simulations') || has(resp.data, 'hybrid_simulations');
```
rstijerina, line 163: We already inspect `resp.data` for `simulations` and `hybrid_simulations` above. Could you set `isSimulation` there?

PR #783, opened 2020-08-12
Fix alert email. Uses a string instead of a list, 'cause of env var fun

rstijerina:
The `NEW_ACCOUNT_ALERT_EMAILS` environment variable was being cast as a string, so we do away with the array and use commas to separate the emails.


designsafe/settings/common_settings.py
```diff
@@ -442,7 +442,7 @@
 EMAIL_HOST_PASSWORD = os.environ.get('SMTP_PASSWORD', '')
 DEFAULT_FROM_EMAIL = os.environ.get('DEFAULT_FROM_EMAIL', 'no-reply@designsafe-ci.org')
 MEETING_REQUEST_EMAIL = os.environ.get('MEETING_REQUEST_EMAIL', 'info@designsafe-ci.org')
-NEW_ACCOUNT_ALERT_EMAILS = os.environ.get('NEW_ACCOUNT_ALERT_EMAILS', ['no-reply@designsafe-ci.org'])
+NEW_ACCOUNT_ALERT_EMAILS = os.environ.get('NEW_ACCOUNT_ALERT_EMAILS', 'no-reply@designsafe-ci.org,')
```
fnets, line 445: Good catch!

PR #789, opened 2020-09-01
Task/DES-1695 - Add formatted metadata to publication archive

Strmiska:
This will create a JSON document to be included with the publication's archive. This will eventually be replaced with the Fedora system once it is up and running again.

I decided to remove the archive method from the Publication class and set it in the Publication Manager. (seems to make more sense to have it there until all of this can be refactored)

**Things to Note:**
- There is a settings name change `DATACITE_URL = settings.DATACITE_URI ---> DATACITE_URL = settings.DATACITE_URL`
- The `zip_publication_files` tasks has been updated with the manager function
- The `archive` function in the publication manager is mostly the same with the biggest change being the addition of the `create_metadata` function within it.

**To Test Locally:**
- You can shell into the django container and create the following paths to simulate the published directories:
`/corral-repl/tacc/NHERI/published/archives/`
`/corral-repl/tacc/NHERI/published/PRJ-XXXX/` - (this prj number will be the name of the project you want to test on)
- use `docker cp` or find some random files to place into the `/corral-repl/tacc/NHERI/published/PRJ-XXXX/` path. This data is what will be archived usually.
- To test, pick a published project (you must have metadata for the publication locally or this will not work)
 - To check if you have the published metadata locally:
```
from designsafe.libs.elasticsearch.docs.publications import BaseESPublication
pub = BaseESPublication(project_id="PRJ-XXXX")
pub.to_dict()
```
- Run the task:
```
from designsafe.apps.api.tasks import zip_publication_files
zip_publication_files('PRJ-XXXX')
```
- Let me know if you see any errors with this as it could be related to something with older or inconsistant project/publication fields.
- You can use `docker cp` to download the zipfile located in `/corral-repl/tacc/NHERI/published/archives/`. Note that this path will have permissions locked down the same way as it is in prod. You can fix this by `chmod`ing the published path in your django container. When the file is unzipped it should contain both a file with the PRJ-ID containing the contents of the published project as well as a PRJ-XXXX_metadata.json file.
- When you're done you can delete the paths you created. Let me know if you have any questions or suggestions!


jarosenb (approved):
I pushed a small commit to prevent errors if the publication doesn't have one of the keys from `entity_type_map`. Otherwise everything worked super smoothly for me locally!

PR #792, opened 2020-09-11
Hotfix: fix dropbox oauth flow and stop unnecessary decoding of external resources file paths

jarosenb:
This is at least one of the things that's breaking transfers in External Resources. These operations depend on having Corral mounted so they can't be tested locally.


designsafe/apps/api/external_resources/box/filemanager/manager.py
```diff
@@ -292,7 +292,7 @@ def download_file(self, box_file_id, download_directory_path):
         """
         box_file = self.box_api.file(box_file_id).get()
         # convert utf-8 chars
```
nathanfranklin, line 294: all these convert comments could be deleted

PR #793, opened 2020-09-16
Bug/DES-1737 - Prevent move and rename of files containing categorized data in projects

Strmiska:
We want to prevent users from moving or renaming files that are categorized to an entity or contain files that are categorized to an entity. If this happens project will become unresponsive and requires manual effort to fix.

This check will only take place when moving/renaming from the My Projects area. With these changes we will request entities for the project and check that the file path(s) to be move/renamed are not among them. To test this feature, go into any project and try to move/copy files. If there are categories associated to the files you should be notified.


jarosenb (approved):
Now that I've experienced having to fix someone's metadata I appreciate how urgent these changes are. I moved your code into the new `checkForEntities` method and it's working for me locally.

PR #794, opened 2020-09-16
Task/DES-1711

owaisj:
This PR does two things.
1. JSON is "pretty printed" in a code block in the Data Depot/Files Preview Modal
2. Valid GeoJSON (`.json` or `.geojson` files) by way of the geojson-validator library is rendered in an instance of a Leaflet map in a modal.

To test, run `npm install` then open a GeoJSON preview modal.

image


nathanfranklin (changes requested):
Looks good 👍  

Left some comments but I'll look when the PR is no-longer-draft.

Two things to look at:
* Fails on some geojson files created from hazmapper v1.  I'll collect and share with you the failing geojson file from hazmapper v1 and look for some other files that might represent hazmapper v1 or v2 plus some edge cases with you.
* I wonder if there are other things we should display like `properties` in a pop up.

jarosenb (approved):
LGTM and Nathan's requested changes seem to be integrated.

nathanfranklin (commented):
@jarosenb , just walked though the recent changes with @owaisj  👍  💯

PR #798, opened 2020-09-30
Task/DES-1752 - Fix publication status update issue with other publications

Strmiska:
Other publications do not have entities, and when they are being published part of the publication task chain will fail silently and prevent the publication status from being set correctly.


designsafe/apps/projects/managers/publication.py
```diff
@@ -196,15 +196,16 @@ def publish_resource(project_id, entity_uuids=None, publish_dois=False):
     responses = []
 
     if publish_dois:
```
jarosenb, line 198: It looks like we can remove an indent level by doing `if publish_dois and entity_uuids:`

jarosenb (changes requested):

jarosenb (approved):
LGTM

PR #803, opened 2020-10-07
DES-1423--user-creation-form

SilversunKSauri:
## Overview: ##
Updating the User Creation form to have all the full user profile fields.

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-1423](https://jira.tacc.utexas.edu/browse/DES-1423)

## Summary of Changes: ##
Refactored the User Creation form to inherit from the User Profile form and the Professional Profile form.
Some pylint fixes.

## Testing Steps: ##
0. Comment out line 390 in forms.py to disable the captcha.
1. Run the branch
2. Go to the "create account" page
3. Confirm that all the fields are there in order
      The proper order is in the field_order value in the form, plus anything not in that value should be in the order it was added to the form.
4. Create a new user.
5. Drop into the django shell and check that the user was created, that the pro profile form was saved, and if you can, check that the user is in Pending state in TAS.

Steps for #5 above:
- docker exec -it des_django /bin/bash
- python manage.py shell
- from designsafe.apps.accounts.forms import ProfessionalProfileForm
- from designsafe.apps.accounts.models import DesignSafeProfile
- me = DesignSafeProfile.objects.get(user_id=###) <- this is the numeric id of the user you created.  you can get to this by logging in to your local portal as your main user, and go into the admin panel and search for your new user and see the id of the user in the url.  This assumes you've either been given admin access in the version of the db you have locally or you have given yourself admin access.
- me_pro = ProfessionalProfileForm(None, instance=me)
- me_pro.fields <- to output the fields and confirm that the form exists


jarosenb (approved):
LGTM!

PR #805, opened 2020-10-09
hotfix/20201009-postit-href: Always list files before creating postits

jarosenb:
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Summary of Changes: ##
Perform a listing before each call to the postits API in order to ensure that the permissions are correct.


fnets (approved):
Works locally and LGTM!

PR #809, opened 2020-10-21
DES-1690, DES-1395: Update Django to 1.11 LTS; Resolve Frontend Security Vulnerabilities

rstijerina:
## Overview: ##
This PR primarily addresses a number of security concerns both in the frontend and the backend via updating packages and removing unused package cruft.

Also removes extraneous committed packages from `designsafe/static`.

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-1690](https://jira.tacc.utexas.edu/browse/DES-1690)
* [DES-1395](https://jira.tacc.utexas.edu/browse/DES-1395)

## Summary of Changes: ##

- commit missing migrations
- update django to 1.11 lts and update related packages
- update dockerfile; only build when debug is false
- add Makefile
- remove unneeded committed packages; use min where possible
- secure translations
- update packages to address security vulnerabilities; linting and webpack config

## Testing Steps: ##
1. run `make build` from the root project directory
2. run `make start`
3. run `npm ci`
4. run `npm run dev`
5. Ensure that all components render properly and all actions execute correctly.


designsafe/apps/api/templates/designsafe/apps/api/agave/preview.html
```diff
@@ -95,8 +95,8 @@
         </p>
     {% endif %}
 
-    <script src="{% static 'vendor/jquery/dist/jquery.js' %}"></script>
-    <script src="{% static 'vendor/bootstrap-ds/js/bootstrap.js' %}"></script>
+    <script src="{% static 'vendor/jquery/dist/jquery.min.js' %}"></script>
```
jarosenb, line 98: I think we can actually delete all of these preview templates. This is handled by Angular now.

jarosenb:
AFAIK the `djng` module actually does need to be in source control. It's an ancient piece of code that's not in NPM. I've deleted it before and had to bring it back because it became impossible to spin up new dev environments without it.

designsafe/apps/api/notifications/migrations/0004_auto_20201019_1920.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2020-10-19 19:20
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
jarosenb, line 8: Where do these migrations come from?

rstijerina:
> AFAIK the `djng` module actually does need to be in source control. It's an ancient piece of code that's not in NPM. I've deleted it before and had to bring it back because it became impossible to spin up new dev environments without it.

It's actually provided by the `django-angular` python package in [requirements.txt](https://github.com/DesignSafe-CI/portal/pull/809/files?file-filters%5B%5D=.txt#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R30). The djng scripts get copied to staticfiles after running `collectstatic`.

package.json
```diff
@@ -6,65 +6,67 @@
...
     "ngtemplate-loader": "^2.0.1",
     "node-sass": "^4.14.1",
-    "sass-loader": "^6.0.7",
-    "style-loader": "^0.20.3",
-    "webpack": "^3.12.0",
-    "webpack-livereload-plugin": "^0.10.0"
+    "prettier": "^2.1.2",
+    "sass-loader": "^8.0.0",
+    "style-loader": "^1.0.1",
+    "webpack": "^5.1.3",
```
jarosenb, line 54: `karma-webpack` doesn't work with webpack 5 (and webpack 4 causes the test runner to throw really cryptic dependency injection errors). I think as long as we're using Angular 1 we're stick with Webpack 3.

designsafe/apps/api/notifications/migrations/0004_auto_20201019_1920.py
rstijerina: I think they came just from running `./manage.py makemigrations`. Might have been from the switch to python3 wrt `bytes` string changes

package.json
rstijerina: Yeah that's a bummer. Seemed like others had been able to get it working with webpack 5, but might have to roll it back if I can't

jarosenb:
> > AFAIK the `djng` module actually does need to be in source control. It's an ancient piece of code that's not in NPM. I've deleted it before and had to bring it back because it became impossible to spin up new dev environments without it.
> 
> It's actually provided by the `django-angular` python package in [requirements.txt](https://github.com/DesignSafe-CI/portal/pull/809/files?file-filters%5B%5D=.txt#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R30). The djng scripts get copied to staticfiles after running `collectstatic`.

Oh word. In that case we still need them checked in if we want to run our pipelines, otherwise we'll need to write a runner that executes a bunch of django code in front of the angular tests. We should plan to phase out djng entirely, since we already have a user service that keeps track of auth state.

package.json
rstijerina: Got it working with v4

rstijerina:
> > > AFAIK the `djng` module actually does need to be in source control. It's an ancient piece of code that's not in NPM. I've deleted it before and had to bring it back because it became impossible to spin up new dev environments without it.
> > 
> > 
> > It's actually provided by the `django-angular` python package in [requirements.txt](https://github.com/DesignSafe-CI/portal/pull/809/files?file-filters%5B%5D=.txt#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R30). The djng scripts get copied to staticfiles after running `collectstatic`.
> 
> Oh word. In that case we still need them checked in if we want to run our pipelines, otherwise we'll need to write a runner that executes a bunch of django code in front of the angular tests. We should plan to phase out djng entirely, since we already have a user service that keeps track of auth state.

Added back the minified file to the source for the purposes of testing.

> We should plan to phase out djng entirely, since we already have a user service that keeps track of auth state.

Totally.

jarosenb (approved):
LGTM now that pipelines are passing.

PR #812, opened 2020-10-23
task/DES-1225 - Google Dataset Search

fnets:
## Overview: ##
Added metadata to publication template for Google Dataset search.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-1225](https://jira.tacc.utexas.edu/browse/DES-1225)

## Summary of Changes: ##
Pulled data from publication model and ElasticSearch, and dumped it into a JSON, which is then put into the template for publication.

## Testing Steps: ##
1. Look at each type of publication, and check the HTML template (cmd + option + u)
2. Look for the tag <script type="application/ld+json">
3. Make sure the data is there, and that it passes a sanity check.
4. Insert publication's URL into Google's rich results test: https://search.google.com/test/rich-results


designsafe/apps/projects/models/agave/base.py
```diff
@@ -263,6 +264,87 @@ def update_archive(project_directory):
...
+        else:
+            related_ents = self.related_entities()
+            logger.debug(related_ents)
+        if getattr(self, 'team_order', False):
+            authors = sorted(self.team_order, key=lambda x: x['order'])
+        else:
+            authors = [{'name': username} for username in [self.pi] + self.co_pis]
+        dataset_json['creator'] = generate_creators(authors)
+        dataset_json['author'] = generate_creators(authors)
+        if BaseESPublication(project_id=self.project_id):
```
jarosenb, line 340: I'd like to phase out the use of `BaseES` classes because they sometimes fail to wrap documents correctly. We can use the `IndexedPublication classes instead.
```suggestion
        from designsafe.apps.data.models.elasticsearch import IndexedPublication, IndexedPublicationLegacy
        from designsafe.libs.elasticsearch.exceptions import DocumentNotFound
        ...
        try:
            pub = IndexedPublication.from_id(self.project_id)
        except DocumentNotFound:
            pass
         try:
            pub = IndexedPublicationLegacy.from_id(self.project_id)
         except DocumentNotFound:
            pass
```

jarosenb (dismissed):
Looks good but I'd like to request a small refactor to remove a dependency on the `BaseESPublication` classes.

designsafe/apps/projects/models/agave/base.py
fnets: Done!  Thanks for the suggestion.

owaisj (approved):
LGTM 👍

PR #813, opened 2020-10-28
DES-661 - Fix user-agent warning for supported IOS browsers.

duckonomy:
## Overview: ##
Fixes an incorrect error message that pops up on supported browsers on IOS devices.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-661](https://jira.tacc.utexas.edu/browse/DES-661)

## Summary of Changes: ##
Added a separate condition for IOS user-agent names (FxiOS, CriOS) for supported browsers. 

## Testing Steps: ##
1. Open the site on supported browsers (Firefox and Chrome) on IOS.
2. Confirm that you do not see the error message about unsupported browsers.

## UI Photos:
N/A

## Notes: ##
I have tested with two IOS devices and my Android phone with both Firefox and Chrome on each device to confirm correct behavior.


designsafe/middleware.py
```diff
@@ -19,13 +19,13 @@
...
     """
-    Middleware to check if the user is running Chrome or Firefox and 
+    Middleware to check if the user is running Chrome or Firefox and
     flash a warning otherwise.
     """
 
     def process_request(self, request):
         user_agent = request.META['HTTP_USER_AGENT']
-        agent_is_supported = ('Chrome' in user_agent) or ('Firefox' in user_agent)
+        agent_is_supported = ('Chrome' in user_agent) or ('Firefox' in user_agent) or ('FxiOS' in user_agent) or ('CriOS' in user_agent)
```
rstijerina, line 28: Now that we have 4 user agent strings, referenced both here and in `base.html`, perhaps we should pull them out into a setting, `SUPPORTED_USER_AGENTS = ['Chrome', 'Firefox', 'FxiOS', 'CriOS'], and this would instead become something like
```suggestion
        agent_is_supported = any(agent in user_agent for agent in settings.SUPPORTED_USER_AGENTS)
```

rstijerina (changes requested):

designsafe/middleware.py
duckonomy: That sounds good!

duckonomy: I'll make changes to settings.

duckonomy: @rstijerina Would there be a simple way to use Django settings in the template as well?

rstijerina: @duckonomy Ah yes, we may want to refactor this a bit. You can reference variables in a django template via `context`, and the [messages](https://docs.djangoproject.com/en/1.11/ref/contrib/messages/) framework has an accompanying context processor in `designsafe/context_processors.py`.
1. Add a context variable to the `messages` function of `context_processors` that  evaluates the `agent_is_supported` logic from `middleware.py` (after looking closer I'm not even sure this is active)
2. [Reference the boolean in the `messages.html` template ](https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#if)
3. Move the html in `base.html` that displays the warning into the conditional in `messages.html`

duckonomy: @rstijerina Thanks for the insight!

jarosenb:
I'm concerned about changing the source of truth for the user agent from the javascript `navigator` to the Django request. If I remember right we had to abandon the user agent middleware because in staging/production it was returning information about the *server* instead of the user.

duckonomy:
@jarosenb Yes, I think Sal also mentioned that the user-agent check in `middleware.py` could be inactive.
Maybe that's the reason why someone chose to use it in JS.
Should I just rollback and use the `navigator`?

duckonomy:
@jarosenb I think it's ready to merge.

jarosenb (approved):
LGTM!

PR #815, opened 2020-11-02
DES-773 - Add message to notify duplicate email in Edit Profile

duckonomy:
## Overview: ##
Added a way to handle the duplicate email error in the interface.

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-773](https://jira.tacc.utexas.edu/browse/DES-773)

## Summary of Changes: ##
Changed `apps/accounts/views.py` to add an error on duplicate emails rather than a silent 500 server error when editing profile.

## Testing Steps: ##
1.Go to the "Edit Profile" page and enter an existing email that is not associated with the currently logged-in account.
2. Confirm that the error message occurs.

## UI Photos:
Screenshot from 2020-10-23 12-39-48

## Notes: ##
I was not sure how to test this given that I was using my other account when testing it for myself.


designsafe/apps/accounts/views.py
```diff
@@ -331,18 +331,27 @@ def profile_edit(request):
 
     if request.method == 'POST':
         form = forms.UserProfileForm(request.POST, initial=tas_user)
+
         if form.is_valid() and pro_form.is_valid():
+            tas_email_exists = tas.get_user(email=form.cleaned_data['email']) and (tas_user['email'] != form.cleaned_data['email'])
```
jarosenb, line 336: Please update the unit tests to check for this case (right now `tas_email_exists` always returns `True` in the test context because mocks are truthy). Feel free to ping me if you want to pair up on this!

jarosenb (changes requested):

designsafe/apps/accounts/views.py
duckonomy: @jarosenb Fixed!

designsafe/apps/accounts/views.py
```diff
@@ -332,6 +332,11 @@ def profile_edit(request):
     if request.method == 'POST':
         form = forms.UserProfileForm(request.POST, initial=tas_user)
         if form.is_valid() and pro_form.is_valid():
+            if (tas.get_user(email=form.cleaned_data['email']) and
```
jarosenb, line 335: This code could be rewritten to make it a little more explicit that we're checking for a different user with the same email. Something like:
```suggestion
            existing_user = (tas.get_user(email=form.cleaned_data['email'])
            if existing_user and existing_user['username'] != tas_user['username]:
                # error logic
```

jarosenb (changes requested):

designsafe/apps/accounts/views.py
duckonomy: @jarosenb  Cool!

duckonomy:
@jarosenb I think this is ready to merge!

fnets (approved):
LGTM, and works locally!  Good work!

jarosenb (approved):
Looks good! Thanks for making those changes.

PR #817, opened 2020-11-05
task/DES-1764: add primary natural hazards interest field

jarosenb:
## Overview: ##
Add a "primary natural hazards interest" selector to the user profile form and make the original checkbox options.


SilversunKSauri (approved):
Looks good to me.  I was able to register a new user with the new fields, and updating profile in the manage account using my primary account worked as expected.

PR #821, opened 2020-11-11
hotfix/warn-trash-entities: Prevent trashing of data with associated entities

jarosenb:
Use the same logic for preventing move/rename to prevent trashing as well.


fnets (approved):
LGTM, works locally, and good refactor!  Good job!

PR #841, opened 2021-01-19
fix Google Drive connection message

rstijerina:
## Overview: ##

This fixes the wrong message was appearing for users when prompted to connect their Google Drive accounts.


jarosenb (approved):
LGTM, thanks for catching this!

PR #858, opened 2021-03-29
Task/DES-1902 - Connect endpoints for versioning and amends

Strmiska:
## Overview: ##
This PR is meant to complete the Amend and Revision pipelines. The process of amending and versioning a project is mostly complete.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-1902](https://jira.tacc.utexas.edu/browse/DES-1902)

## Testing Steps: ##
**NOTE: File copying and directory creation is not currently testable from the Dev environment. This may be possible to test in Staging though. If you plan to submit any dev project for publication, revision, or amendments please DM me on Slack with the correct designsafe.env settings to use. I'm open to have some group testing on this**
1. Navigate into an "other" project that has files.
2. Go to the "Prepare to Publish" section.
 - From this section you should be able to see options for revise or amend if your "other" project is published. If it is not, you can still access these through the browser console by running the `goVersion()` and `goAmend()` functions. Once you navigate you should be able to test the various sections of the pipeline.

## UI Photos:

Screen Shot 2021-03-28 at 8 23 58 PM
Screen Shot 2021-03-30 at 6 12 00 PM
Screen Shot 2021-03-30 at 6 12 14 PM
Screen Shot 2021-03-30 at 6 14 32 PM
Screen Shot 2021-03-30 at 6 12 51 PM

### DIFF
--- a/designsafe/apps/projects/managers/publication.py
+++ b/designsafe/apps/projects/managers/publication.py
@@ -233,6 +233,10 @@ def amend_datacite_doi(publication):
     
     :param elasticsearch publication: Publication to amend
     """
+    # Only amend doi while in prod
+    if getattr(settings, 'DESIGNSAFE_ENVIRONMENT', 'dev') != 'default':
+        return
+
     pub_dict = publication.to_dict()
     prj_class = project_lookup_model(pub_dict['project'])
     project = prj_class(value=pub_dict['project']['value'], uuid=pub_dict['project']['uuid'])
--- a/designsafe/apps/projects/managers/publication.py
+++ b/designsafe/apps/projects/managers/publication.py
@@ -17,6 +17,7 @@ from designsafe.apps.api.agave import service_account
 from designsafe.apps.data.models.agave.files import BaseFileResource
 from designsafe.apps.projects.managers import datacite as DataciteManager
 from designsafe.apps.projects.managers.base import ProjectsManager
+from designsafe.apps.projects.models.utils import lookup_model as project_lookup_model
 from designsafe.libs.elasticsearch.docs.publications import BaseESPublication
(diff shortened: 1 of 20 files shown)


designsafe/apps/projects/managers/publication.py
```diff
@@ -178,6 +179,72 @@ def draft_publication(
...
+    }
+    for key in amends_dict:
+        amends_dict[key] = prj_dict['value'][key]
+
+    pub_dict['project']['value'].update(amends_dict)
+    pub.update(**pub_dict)
+
+    return pub
+
+def amend_datacite_doi(publication):
```
jarosenb, line 226: Should we make this method prod-only so that we can test it in dev/staging without making real datacite calls?

Strmiska: That's a good idea, I could set it up to use the test API that DataCite has, but we'd need to change some environment variables to include that. If I check for the `DESIGNSAFE_ENVIRONMENT` we could switch the  `DATACITE_USER` to a  `TEST_DATACITE_USER`... I'm not too familiar with how that looks on the staging/prod side of things though.
https://github.com/DesignSafe-CI/portal/blob/master/designsafe/apps/projects/managers/datacite.py#L16

jarosenb: If we're confident in how we're formatting the calls then the best practice might be to return early in non-prod environments, but include good unit tests. I'd be happy to submit a supplemental PR to this one with test coverage.

Strmiska: Okay, I'll just have it return early if we're not in the prod environment.

PR #883, opened 2021-06-02
Task/DES-1976 - Fix missing file tags for versioned publications

Strmiska:
## Overview: ##
Subsequent versions of a publication would not correctly update the file uuids for the file tags in a project.
We needed to adjust a regex pattern in the file listing for publications as well as update a base path being used when attempting to correct the file uuid being updated in the publication/versioning process.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-1976](https://jira.tacc.utexas.edu/browse/DES-1976)

## Testing Steps: ##
1. You'll have to copy your project files into a versioned directory you want to test on corral. For example: you would copy the files (which have tags in your project to /published/PRJ-1234v2)
2. When those files are copied shell into the django container running locally and run the following task 
`from designsafe.apps.api import tasks` 
`tasks.swap_file_tag_uuids('PRJ-1234', revision=2)`
3. Check your publication to make sure the file tags appear (in this case v2 of your publication)

### DIFF
--- a/designsafe/apps/projects/managers/publication.py
+++ b/designsafe/apps/projects/managers/publication.py
@@ -495,16 +495,11 @@ def fix_file_tags(project_id, revision=None):
                     fix_tags_no_path(entity)
         else:
             if 'value' in pub_dict[entname] and 'fileTags' in pub_dict[entname]['value'] and check_complete_tags(pub_dict[entname]['value']['fileTags']):
-                logger.info('FIRST IF')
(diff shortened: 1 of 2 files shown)


jarosenb (approved):
LGTM, post site visit we should revisit this and try to deprecate `BaseFileResource`.

PR #884, opened 2021-06-02
task/DES-1963 - Curation Directory mentions Categories for Other

fnets:
## Overview: ##
Other-type projects don't have user-definied categories, so the tutorial explaining them was removed from other-type projects and replaced with a tutorial to move files.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-1963](https://jira.tacc.utexas.edu/browse/DES-1963)

## Summary of Changes: ##
The template checks if the project's category is other, and displays the correct tutorial as needed.

## Testing Steps: ##
1. Create an empty other-type project or DM Frank to be added to one.
2. Check that the "Working directory" and "Curation directory" both display the tutorial for help moving files, and do not display the tutorial for creating categories
3. Create any other type of empty project
4. Check that the "Curation directory" displays the tutorial for creating categories and that the "Working directory" displays the tutorial for help moving files

## UI Photos:
Curation Directory for an empty other-type project:
 Screen Shot 2021-06-02 at 4 20 40 PM


designsafe/static/scripts/data-depot/components/data-depot-listing/files-listing.template.html
```diff
@@ -143,11 +143,19 @@
...
                     </p>
-                    <p class="lead" style="margin-top:20px" ng-if="$ctrl.editTags">
+                    <p class="lead" style="margin-top:20px" ng-if="$ctrl.editTags && !$ctrl.getType(listing.params.system) == 'Other'">
                         <i class="fa fa-file-text-o">&nbsp;</i><br/>
                         <a href="https://www.youtube.com/watch?v=lfBe04ihQUs&list=PL2GxvrdFrBlkwHBgQ47pZO-77ZLrJKYHV" target="_blank">
                             <i class="fa fa-question-circle" style="font-size: 18px;">&nbsp;</i>
                             Learn how to assign files to categories<br/>
+                        </a>  
+                    </p>
+                    <p class="lead" style="margin-top:20px" ng-if="$ctrl.editTags && $ctrl.getType(listing.params.system) == 'Other'">
```
Strmiska, line 153: `listing.params.system` will not have the project type you're looking for regarding these projects. I don't actually think the project model is anywhere in the listing at the moment. We'll have to dig into the data bindings being passed into the FileLisingComponent. You'll see some one way data bindings that look like `showTags` or `editTags` This might be one way we could set a parameter like `missingFileMessage` that could be a bit more dynamic. Might discuss this in the next meeting.

Strmiska (changes requested):
We'll have to rethink how we do this and take some things into consideration. This will require a more involved fix.

PR #893, opened 2021-06-24
task/DES-2019: Replace node-sass dev dependency with sass

jarosenb:
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

## Testing Steps: ##
1.


rstijerina:
Out of curiosity, what was this change predicated on?

jarosenb:
> Out of curiosity, what was this change predicated on?

A couple reasons:
- The `node-sass` library is deprecated and this is the suggested migration path: https://sass-lang.com/blog/libsass-is-deprecated
- When node v16 becomes the LTS version and people start using it locally, it will break stuff because the `node-sass` version in package.json needs to be synced with the system-wide node install.

rstijerina (approved):

PR #896, opened 2021-07-02
Task/DES-2013 - Fix issues with collapse tab target on published entities

Strmiska:
## Overview: ##
This fixes a bug related to the collapsing entity tabs on published projects being frozen or not opening when clicked (under certain conditions). Because an entity can be used or categorized multiple times within a publication template we can't use just the `entity.uuid` as the target. We need to account for the possibility of an entity being reused at any level within the project structure. I've attached a simple diagram to explain what I mean.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2013](https://jira.tacc.utexas.edu/browse/DES-2013)

## Testing Steps: ##
1. Before you pull this branch you can use an existing project that has entities (experimental, simulation, field research etc...) create an entity to categorize to your mission, experiment, simulation. Use that same entity in two different missions, experitments etc... You should be able to see the bug in the publication preview section when you try to collapse that entity in the preview
2. pull this PR and check that does not happen. The process works the same across all project types besides "other" (because it does not have any categories)

## UI Photos:
dropdown fix


jarosenb (approved):
LGTM

PR #925, opened 2021-11-29
bug/DES-2128: Project serializers are non-deterministic

jarosenb:
## Overview: ##
We have a bunch of serializers that use `list(set(arr))` to dedupe arrays, which is really bad since that doesn't preserve their order and makes testing extremely difficult.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2128](https://jira.tacc.utexas.edu/browse/DES-2128)

## Testing Steps: ##
The new test suite should provide proof that we're serializing things deterministically now, but testers should click around My Projects to make sure nothing breaks.


Strmiska (approved):
LTGM

PR #930, opened 2021-12-14
Replace "start date" filters with simple show/hide history toggle.

thompscs:
## Overview: ##

Users were confused by the default view and could not easily figure out
how to see what is happening right now in the facilities.  They were being
presented by default with the full history of all projects in all facilities,
sorted from newest to oldest, causing projects scheduled a year or more into
the future to be the first thing displayed when they load the page.

This change replaces the menu of "start date"-based filters entirely with 
a simple checkbox to show or hide projects that are already completed.  
It also reverses the default ordering from "newest first" to "oldest first".

Applying the new "hide history" filter and reversing the sort order by default 
on page load will mean users are now presented immediately with a view
of projects that is more in-line with expectations:

 * Projects in facilities currently are the first things they see, and
 * Upcoming projects appear the further down the page they go.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

## Summary of Changes: ##

* Remove "start date" filter menu from filter bar component.
* Add checkbox to show or hide events with end date in the past.
* Add a default filter to the initial loadProjects() call when the page loads that activates this "hide history" filter.
* Reverse the default sort order from descending by "start date" to ascending.

## Testing Steps: ##
1. Load the app page (/nco)
2. Confirm the events are sorted by oldest "start date" first instead of newest upon initial page load.
3. Confirm the default view on initial page load is not showing events with an "end date" before today.
4. Confirm the show/hide history checkbox is working.

## UI Photos: ##

nco-app-history-on

nco-app-history-off

## Notes: ##

Events that have an empty "start date" field are being sorted at the top of the list.  This is not a code problem.  This is a data problem fixed in the database.  These entries just need to be updated with a start date to fall into their appropriate place in the list.

The code for how to setup a checkbox like this was borrowed from how the Data Depot file listings are doing it.

### DIFF
--- a/designsafe/apps/nco/managers/projects.py
+++ b/designsafe/apps/nco/managers/projects.py
@@ -68,7 +68,13 @@ class NcoProjectsManager(object):
                 return {"dateStart": {"$gt": start, "$lt": end}}
             return in_range
 
+        def no_history():
+            """Hide already completed events."""
+            yesterday = date - timedelta(days=1)
+            return {"dateEnd": {"$gt": yesterday}}
+
         range_switch = {
+            "No History": no_history,
             "Happening This Week": this_week,
             "Happening This Month": this_month,
             "Event Started Last 7 Days": last_n_days(7),
--- a/designsafe/static/scripts/nco/components/nco-scheduler-filters/nco-scheduler-filters.component.js
+++ b/designsafe/static/scripts/nco/components/nco-scheduler-filters/nco-scheduler-filters.component.js
@@ -18,7 +18,7 @@ class NcoSchedulerFiltersCtrl {
             'Start Date - oldest first',
             'Award Number',
             'Facility',
-	    'PI Name'
+            'PI Name'
         ];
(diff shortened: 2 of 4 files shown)


rstijerina (approved):
LGTM!

> Events that have an empty "start date" field are being sorted at the top of the list. This is not a code problem. This is a data problem fixed in the database. These entries just need to be updated with a start date to fall into their appropriate place in the list.

@thompscs , is the missing start data on the database side being fixed?

fnets (approved):
LGTM!

SilversunKSauri (approved):
LGTM

thompscs:
> @thompscs , is the missing start data on the database side being fixed?

This will be an ongoing issue.  We will correct the data we currently have before making a big public announcement about the page, but we will have to keep the facilities' feet to the fire when it comes to providing us complete data about their projects with each update.

PR #934, opened 2021-12-20
Task/DES-1483: Display Author Modal for PIs and Co-PIs

owaisj:
## Overview: ##

Allow users to view author modals for PIs and Co-PIs

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-1483](https://jira.tacc.utexas.edu/browse/DES-1483)

## UI Photos:

image

image

image

### DIFF
--- a/designsafe/static/scripts/data-depot/components/published/published-view.component.js
+++ b/designsafe/static/scripts/data-depot/components/published/published-view.component.js
@@ -19,6 +19,14 @@ class PublishedViewCtrl {
         this.FileOperationService = FileOperationService;
         this.PublicationService= PublicationService;
         this.UserService = UserService;
+        this.loadingUserData = {
+            pi: true,
+            coPis: true,
+        };
+        this.authorData = {
+            pi: {},
+            coPis: null,
+        };
     }
 
     $onInit() {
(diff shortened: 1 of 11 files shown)


designsafe/static/scripts/data-depot/components/published/published-view.component.js
```diff
@@ -19,6 +19,14 @@ class PublishedViewCtrl {
...
         this.PublicationService= PublicationService;
         this.UserService = UserService;
+        this.loadingUserData = {
+            pi: true,
+            coPis: true,
+        };
+        this.authorData = {
+            pi: {},
+            coPis: null,
+        };
```
Strmiska, line 29: Will it cause any issues to initialize these variables in `$onInit` instead of the constructor? If so disregard. I just wanted to mention that `this.ui` is where most of the UI params are set across different components on DS (usually) for things like "loading" etc...

designsafe/static/scripts/data-depot/components/published/published-view.component.js
```diff
@@ -35,6 +43,33 @@ class PublishedViewCtrl {
...
+                email: res.email,
+                name: res.username,
+                inst: res.profile.institution,
+            };
+            this.loadingUserData.pi = false;
+        });
+        if (this.project.value.coPis) {
+            this.authorData.coPis = new Array(this.project.value.coPis.length);
+            this.project.value.coPis.forEach((coPi, idx) => {
+                this.UserService.get(coPi).then((res) => {
```
Strmiska, line 60: I believe there is a working method in the UserService for getting multiple user profiles at once called `getPublic` which should be able to accept a list of usernames.

Strmiska: Oh lame! Yeah if the backend has that data and it's just not being passed up, see if you can add it in to the response. I'm sure it'll come in handy in the future. Sometimes we have like 20 or so users to look up on these projects and I don't really know how long it would take to grab each one.

If it turns out to be a bunch of extra work, just make a new JIRA task for it with the files and line numbers, then we'll come back to it.

PR #935, opened 2021-12-21
DES-2047 Fix: Edit Project url location fix

SilversunKSauri:
## Overview: ##
Moved the remaining Edit Project urls on other templates that I missed before.

Also removed the Edit Project urls from the Publication Preview as per a new secondary request from Craig.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2047](https://jira.tacc.utexas.edu/browse/DES-2047)

## Testing Steps: ##
1. Click through each of the tabs in multiple types of projects.
2. Confirm that the Edit Project is now right-aligned on the Working Directory and Curation Directory pages.
3. Confirm that the Edit Project url does not exist on the Publication Preview page.

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/curation-directory/curation-directory.component.html
+++ b/designsafe/static/scripts/data-depot/components/projects/curation-directory/curation-directory.component.html
@@ -4,8 +4,8 @@
         <div class="prj-head-container">
             <span class="prj-head-title">
                 {{$ctrl.browser.project.value.projectId}} | <strong>{{ $ctrl.browser.project.value.title }}</strong>
-                <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
             </span>
+            <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
         </div>
         <table style="table-layout: fixed; width: 100%; margin-top: 10px; margin-bottom: 10px;">
             <colgroup>
--- a/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
+++ b/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
@@ -73,7 +73,6 @@
         <div class="prj-head-container">
             <span class="prj-head-title">
                 {{$ctrl.project.value.projectId}} | <strong>{{ $ctrl.project.value.title }}</strong>
-                <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
             </span>
         </div>
         <table style="table-layout: fixed; width: 100%; margin-top: 10px; margin-bottom: 10px;">
--- a/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
+++ b/designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
@@ -74,6 +74,7 @@
             <span class="prj-head-title">
                 {{$ctrl.project.value.projectId}} | <strong>{{ $ctrl.project.value.title }}</strong>
             </span>
+            <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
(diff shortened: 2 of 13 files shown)


designsafe/static/scripts/data-depot/components/projects/curation-directory/curation-directory.component.html
```diff
@@ -4,8 +4,8 @@
         <div class="prj-head-container">
             <span class="prj-head-title">
                 {{$ctrl.browser.project.value.projectId}} | <strong>{{ $ctrl.browser.project.value.title }}</strong>
-                <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
             </span>
+            <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
```
Strmiska, line 8: In the components for each publication preview we can remove the `manageProject()` functions that call those modals since we're not using them anymore.

designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview.component.html
```diff
@@ -4,7 +4,6 @@
         <div class="prj-head-container">
             <span class="prj-head-title">
                 {{$ctrl.browser.project.value.projectId}} | <strong>{{ $ctrl.browser.project.value.title }}</strong>
-                <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)" ng-if="!$ctrl.readOnly"><strong>Edit Project</strong></a>
             </span>
```
Strmiska, line 7: In the components for each publication preview we can remove the `manageProject()` functions that call those modals since we're not using them anymore.

designsafe/static/scripts/data-depot/components/projects/curation-directory/curation-directory.component.html
Strmiska, line 8: I'm kinda confused as to why I see this change only for the `curation-directory` template and not the `project-view` template. I'm assuming that the Edit Project button would be in two different places before, but apparently this change is already made in the `project-view`?

Strmiska: I think the only other places that need this change are in the following templates:
`pipeline-amend.template.html` - this is in the "Amend" pipeline for other projects that have been published. I can show you how to get to this page if you don't have an "other" project to use
`pipeline-project.template.html` - This is in the normal publication pipeline after the selection step.

SilversunKSauri: the change for the `project-view` template was already merged in, it's in staging currently if you want to check.
I'll check those other templates and look at the other removals you mentioned.

SilversunKSauri: Do we need to not allow a user to do an "edit project metadata" from the `pipeline-amend.template.hmtl`?  When I did these updates, I did a search for "edit project" and I saw this one, but I think it's more of amending the information for the project metadata, which I think should be allowed?

It does look like I somehow missed `pipeline-project.template.html`, so updating that file.

designsafe/static/scripts/data-depot/components/projects/pipeline-project/pipeline-project.component.html
```diff
@@ -73,7 +73,6 @@ <h3 ng-if="$ctrl.project.value.projectType != 'other'" style="text-align: center
         <div class="prj-head-container">
             <span class="prj-head-title">
                 {{$ctrl.project.value.projectId}} | <strong>{{ $ctrl.project.value.title }}</strong>
-                <a class="prj-head-edit" ng-click="$ctrl.manageProject($event)"><strong>Edit Project</strong></a>
             </span>
```
Strmiska, line 76: Unless Craig and Maria said to remove this here, we should Keep the Edit Project button and modal in place here. I would just add the formatting changes you had earlier so that it's right aligned.

Strmiska (approved):
LGTM

PR #948, opened 2022-02-28
task/DES-2203: Port CEPV2 file indexer to Designsafe

jarosenb:
## Overview: ##
Ports the CEPV2 file indexer into Designsafe. This gives us bulk operations and fast lookup by system/path. There should now be ~99% fewer network calls made during indexing, which will hopefully prevent our queues from crashing so frequently.

The plan for deployment is this: 
1. Create a new "reindexing" index to copy our existing documents into
2. Copy existing documents and give each one a new-style UUID (using the uuid_fix utility method).
3. Perform an atomic alias swap to point future search operations at the update index.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2203](https://jira.tacc.utexas.edu/browse/DES-2203)

## Testing Steps: ##
TBD


Strmiska (approved):
Very nice, LGTM! Let me know if we can or need to coordinate anything for testing.

PR #951, opened 2022-03-02
task/DES-2111: Add ds-use-case docs container

rstijerina:
## Overview: ##

- Adds `designsafe_use_case` container, with watchtower enabled look for updates to `designsafeci/ds-use-case-template`
- Adds flags `WATCHTOWER_INCLUDE_STOPPED` and `WATCHTOWER_REVIVE_STOPPED` to enable watchtower for the stopped `designsafe_use_case` container, as it exits after build

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2111](https://jira.tacc.utexas.edu/browse/DES-2111)


jarosenb (approved):
Works in ds-next

PR #969, opened 2022-04-28
task/DES-2241: Add footerless template; add back nav and footer padding

rstijerina:
## Overview: ##

- DS Uses Cases group does not want footer on page, so we need to create a new CMS template without a footer.
- Also per Hedda, margin from [DES-2206](https://github.com/DesignSafe-CI/portal/pull/954) was intentional, so revert that change from main template.

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2241](https://jira.tacc.utexas.edu/browse/DES-2241)
* [DES-2206](https://jira.tacc.utexas.edu/browse/DES-2206)


jarosenb (approved):
LGTM, I thought we'd have to do something way way uglier to get django-cms to kill the footer on a specific page.

PR #971, opened 2022-05-02
task/DES-2248: Allow apps with no metadata record

rstijerina:
## Overview: ##
Simcenter wants to be able to launch public apps not in public tray. This change makes it more like other portals, where if a valid app id is supplied in the URL, the app will be loaded.

To test, try the public app `quoFEM-frontera-1.0.0u2` which has no metadata record

## PR Status: ##

* [X] Ready.
## Related Jira tickets: ##

* [DES-2248](https://jira.tacc.utexas.edu/browse/DES-2248)

## Testing Steps: ##
1. Go to https://designsafe.dev/rw/workspace/#!/quoFEM-frontera-1.0.0u2
2. Confirm app is loaded

## UI Photos:
Screen Shot 2022-05-02 at 2 46 25 PM

### DIFF
--- a/designsafe/static/scripts/workspace/components/application-tray/application-tray.component.js
+++ b/designsafe/static/scripts/workspace/components/application-tray/application-tray.component.js
@@ -101,10 +101,15 @@ class AppTrayCtrl {
                                         .parent($('#toast-container')));
                                 }
                             } else {
-                                this.$mdToast.show(this.$mdToast.simple()
-                                    .content(this.$translate.instant('error_app_run'))
-                                    .toastClass('warning')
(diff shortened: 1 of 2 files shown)


jarosenb:
LGTM, but just to clarify, does this mean that there's no need for these metadata-less apps to show up in the tray?

rstijerina:
> LGTM, but just to clarify, does this mean that there's no need for these metadata-less apps to show up in the tray?

Exactly- they want some apps to be accessible via URL, but not be shown in the apps tray.

jarosenb (approved):
LGTM

PR #972, opened 2022-05-02
task/DES-2116 - Added new metadata for Google Datasets

fnets:
It should now recognize the related entities of publications and be cleaner.

## Overview: ##
Google Data Search is indexing only at the Project level. Instead, datasets with dois at the experiment/mission/sim levels are not being indexed. We are mostly interested in marking up our cited publications with DOI.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2116](https://jira.tacc.utexas.edu/browse/DES-2116)

## Summary of Changes: ##
Nested related entities in the publications' JSON+LD for Google Dataset. They should provide all the information that the publication does, plus their dois

## Testing Steps: ##
1. Navigate to any publication with related entities (ie. Experimental publications)
2. View Page Source (CMD+OPTION+U)
3. Find `<script type="application/ld+json">`
4. Copy/paste JSON into the Rich Results test:  https://search.google.com/test/rich-results
5. Make sure that `relatedIdentifier_X` appears (where X is an integer) and the entry generally makes sense.

## Notes: ##
The most thorough test for this will be releasing it and seeing how Google Dataset picks up on it.


designsafe/apps/projects/models/agave/base.py
```diff
@@ -265,40 +265,134 @@ def to_dataset_json(self):
         }
         if self.dois:
```
Strmiska, line 266: Can this be moved under the same `if self.dois:` on L328?

Strmiska (approved):
LGTM We'll need to make a new task for the custom "Other" fields for each project type.

PR #978, opened 2022-05-16
DES-2242: Publication Pipeline File Check

SilversunKSauri:
## Overview: ##
Added new celery task at the end of the Publication Pipeline to check for missing files/folders and empty top-level folders.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##
* [DES-2242](https://jira.tacc.utexas.edu/browse/DES-2242)
* [DES-2169](https://jira.tacc.utexas.edu/browse/DES-2169) - original request, resource issues means this exact task won't be done, but 2242 is the current solution for now.

## Summary of Changes: ##
Added new celery task to check for missing files/folders and empty top-level folders.
Specific devs will get an email if something is missing or empty.
NOTE: This is not a comprehensive "going down the tree of folders" check, only a single directory down so we don't use up too many resources traversing the folder structure.

## Testing Steps: ##
1. Make sure you have a publication in your local elasticsearch and copy the files to your local version of the publication path.  PRJ-2224 has a small number of files and is a dev published project, so that's a good candidate.  (Let me know if you have trouble with this step and I can show how to do this.)
2. Go to common_settings.py and modify `DEV_PROJECT_ADMINS_EMAIL` to only show your email so you don't accidentally send testing emails to the devs listed.
3. Make sure you are on the VPN, otherwise the email portion will not work.
4. Run the code, exec into django container, run the python shell.
5. Import the celery task, set a project_id variable, then run the celery task on that project_id.
6. If you get no logging messages in the python shell, then that means you have no missing files or empty folders compared to the files/folders you have locally.  Delete a file or empty a folder and run it again.  You should get logging messages and an email.

### DIFF
--- a/designsafe/apps/api/tasks.py
+++ b/designsafe/apps/api/tasks.py
@@ -16,6 +16,7 @@ from designsafe.apps.api.notifications.models import Notification, Broadcast
 from designsafe.apps.api.agave import get_service_account_client
 from designsafe.apps.projects.models.elasticsearch import IndexedProject
 from designsafe.apps.data.models.elasticsearch import IndexedPublication
+from designsafe.libs.elasticsearch.docs.publications import BaseESPublication
 from designsafe.libs.elasticsearch.utils import new_es_client
 from designsafe.apps.data.tasks import agave_indexer
 from elasticsearch_dsl.query import Q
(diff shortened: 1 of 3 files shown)


designsafe/apps/api/tasks.py
```diff
@@ -885,3 +884,86 @@ def email_user_publication_request_confirmation(self, username):
             [user.email],
             html_message=email_body
         )
+
```
Strmiska, line 887: We should add the shared task decorator to this task so that it will run on the default queue. Something like: `@shared_task(bind=True, max_retries=1, default_retry_delay=60)`

SilversunKSauri: Added

Strmiska (approved):
LGTM!

PR #981, opened 2022-06-09
DES-2258: Registration Page Updates

SilversunKSauri:
## Overview: ##
Remove Register form and replace with text and link to TUP

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2258](https://jira.tacc.utexas.edu/browse/DES-2258)

## Summary of Changes: ##
Removing the registration page form and replacing it with text and a link to send users to create new accounts on TUP rather than through DesignSafe. 
Removed some text regarding email confirmations from the password reset page.

## Testing Steps: ##
1.  View the account registration page at /accounts/register and confirm that the link to TUP exists and you don't see the form anymore.
2. Confirm that new users are prompted to fill out their account profile on first login.  This code already existed, but the task had a note to confirm that this was still working.

### DIFF
--- a/designsafe/apps/accounts/templates/designsafe/apps/accounts/password_reset.html
+++ b/designsafe/apps/accounts/templates/designsafe/apps/accounts/password_reset.html
@@ -17,13 +17,6 @@
         <em>as a whole</em>, not just at DesignSafe.
     </div>
 
-    <div class="alert alert-warning">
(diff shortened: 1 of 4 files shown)


fnets (approved):
LGTM! The account profile form pops up on a user's first log in, too. The only correction I have is that the registration url is `account/register/`

PR #1016, opened 2022-11-29
hotfix: Licenses API Endpoint picking up incorrect account to check for staff permissions

edmondsgarrett:
## Overview: ##
Resolved variable name conflict between license API endpoint view and JWT auth
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2330](https://jira.tacc.utexas.edu/browse/DES-2330)

## Summary of Changes: ##
- Changed url param for user to get licenses for to not conflict with JWT auth url param
- Changed default value for JWT auth url user lookup to user associated with included JWT

## Testing Steps: ##
1. Use the JWT provided independently for testing (in these testing steps, it is set as environment variable `JWT_TEST`)
2. Use the following cURL commands and confirm you get back a 200 response and a license for the first and no license for the second (i.e. `{'license': ''}`):
- `curl -i -H "X-Jwt-Assertion-Designsafe: $JWT_TEST" "https://designsafe.dev/api/licenses/MATLAB/?username=gedmonds"`
- `curl -i -H "X-Jwt-Assertion-Designsafe: $JWT_TEST" "https://designsafe.dev/api/licenses/MATLAB/?username=gedtest2"`

### DIFF
--- a/designsafe/apps/api/decorators.py
+++ b/designsafe/apps/api/decorators.py
@@ -93,7 +93,7 @@ def agave_jwt_login(func):
 
         jwt_username = jwt_payload.get(settings.AGAVE_JWT_USER_CLAIM_FIELD, '')
         if jwt_username == settings.AGAVE_JWT_SERVICE_ACCOUNT:
-            username = request.GET.get('user', None)
+            username = request.GET.get('user', jwt_username)
         else:
             username = jwt_username
 
--- a/designsafe/apps/api/licenses/views.py
+++ b/designsafe/apps/api/licenses/views.py
@@ -6,6 +6,9 @@ from django.http.response import HttpResponseForbidden, HttpResponseNotFound
 from django.http import JsonResponse
 from django.apps import apps
 from designsafe.apps.data.models.agave.util import AgaveJSONEncoder
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class LicenseView(SecureMixin, BaseApiView):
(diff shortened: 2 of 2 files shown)


jarosenb (approved):
LGTM on the Designsafe side. We'll need the Jupyterhub spawner to update the query string on its end in order to test.

designsafe/apps/api/decorators.py
```diff
@@ -93,7 +93,7 @@ def decorated_function(request, *args, **kwargs):
 
         jwt_username = jwt_payload.get(settings.AGAVE_JWT_USER_CLAIM_FIELD, '')
         if jwt_username == settings.AGAVE_JWT_SERVICE_ACCOUNT:
-            username = request.GET.get('user', None)
+            username = request.GET.get('user', jwt_username)
```
rstijerina, line 96: What's the rationale behind setting a default `username` value?

edmondsgarrett: With how this if-else is set up, the service account being authenticated with the JWT via request header would not be used to actually make the request, so this change allows that to happen if another user via param `user` is not supplied.

rstijerina: Ahh, yes that makes sense. Good catch.

rstijerina: I wonder if in this case we'd prefer to not actually "login" the `AGAVE_JWT_SERVICE_ACCOUNT` user, since that user's token isn't actually being used but just verified, and instead skip the login phase if no `user` is supplied in this if statement

rstijerina: Going to approve, but curious if anyone has thoughts on this too

rstijerina (approved):

PR #1028, opened 2023-03-20
Task/des 2441 change published search layout

van-go:
## Overview: ##
Change the layout to top, right side, keep columns
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2441](https://jira.tacc.utexas.edu/browse/DES-2441)
_This change is a subtask of_ [DES-2109](https://jira.tacc.utexas.edu/browse/DES-2109)

## Summary of Changes: ##
On the Published page, when a user searched, I removed the show more/less option and instead moved all those options to its own right-nav bar that is always present and available. 

## Testing Steps: ##
1. Navigate to the Published tab
2. Search using different options
3. Make sure that Type options are only available when the type is checked. 
4. Make sure the results are what you'd expect. 

## UI Photos:
Screen Shot 2023-03-20 at 11 37 47 AM

## Notes: ##
The "publication advanced search" options were in their own template and controller. I moved these to the "publications listing' template and controller since are not longer an option, but instead are always available on the page. I did not remove or make changes to publication-advanced-search.template.html  or publication-advanced-search.component.js. Instead, they're just not used. I didn't want to delete anything ... yet.

### DIFF
--- a/designsafe/static/scripts/ng-designsafe/directives/templates/prj-pub-citation-template.html
+++ b/designsafe/static/scripts/ng-designsafe/directives/templates/prj-pub-citation-template.html
@@ -1,0 +1,50 @@
+<div style="background: #D9EDF7;">
+    <div class="entity-meta-data">
+        Cite this Data {{entity}}
+    </div>
+    <div ng-if="$ctrl.selectedVersion === 1"> 
+        <span ng-repeat="author in $ctrl.publication.authors">
+            <span ng-if=$first>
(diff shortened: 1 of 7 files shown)


designsafe/static/scripts/ng-designsafe/directives/templates/prj-pub-citation-template.html
SilversunKSauri, line 1: Is this file relevant to this PR?  might remove if not

PR #1035, opened 2023-04-14
Bug/breadcrumb versioning

van-go:
## Overview: ##
* Breadcrumbs Shows a V2, but V2 doesn’t exist for most projects. 
* Versioned Projects dropdown is showing blank even though it’s going to V2. 
* ‘Version Changes’ shows for V1 even though it should not. 

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

## Summary of Changes: ##
The breadcrumbs component was reliant on prjBasePath. In the PublishedView component, prjBasePath was initially set using revision. I changed it to version, which caused the problem. So I changed it back. 

## Testing Steps: ##
1. Any non-versioned project should show the ProjectId (no-version #) in the breadcrumb. 
2. Any versioned project should show the ProjectId + v# in the breadbrumb. It should also show to the latest version in the dropdown. 
3. When viewing a versioned project, the version dropdown should have 'Version Changes' that displays the date and reason for changed for all version _except_ for version 1. 

## UI Photos:
Screen Shot 2023-04-14 at 9 54 21 AM


jarosenb (approved):
Good catch!

SilversunKSauri (approved):
Did a quick local test, it looks like everything is good on the publications I've been checking

PR #1044, opened 2023-05-24
task/2474 - Google Dataset Compatibility

fnets:
## Overview: ##
Added audits to description field to make sure they're compatible with Google Dataset. 

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2474](https://jira.tacc.utexas.edu/browse/DES-2474)

## Summary of Changes: ##
All description fields should be limited to between 50 and 5000 characters. Added an error message and instructions for the user. Involves a field audiy using Angularjs, and a message that is shown and hidden depending on whether it passes or fails. Also has a new CSS class

## Testing Steps: ##
1. Go to a DesignSafe project
2. Edit the description to be less than 50 characters and then more than 5000 characters, check to make sure you are unable to save the change and that an error message is displayed. (see UI photo)
3. Add or edit an entity (experiement/simulation/etc).  Edit the description to be less than 50 characters and then more than 5000 characters, check to make sure you are unable to save the change and that an error message is displayed. (similar photo as above)
4. Amend a DesignSafe publication. Edit the description to be less than 50 characters and then more than  5000 characters, check to make sure you are unable to save the change and that an error message is displayed.(similar photo as above)
5. Version a DesignSafe publication. Edit the description to be less than 50 characters and then more than  5000 characters, check to make sure you are unable to save the change and that an error message is displayed.(similar photo as above)
6. If there are other places in the publication pipeline where a description is used and are un-audtied, please let me know.

## UI Photos:
Screen Shot 2023-05-24 at 2 07 18 PM

### DIFF
--- a/designsafe/static/scripts/projects/components/amend-entity-modal/amend-entity.template.html
+++ b/designsafe/static/scripts/projects/components/amend-entity-modal/amend-entity.template.html
@@ -90,9 +90,14 @@
                 <textarea class="form-textbox"
                           id="id-description"
                           ng-model="$ctrl.form.description"
-                          ng-required="false">
+                          ng-required="false"
+                          ng-minlength="50"
+                          ng-maxlength="5000">
                     &nbsp;
                 </textarea>
+                <span ng-if="projForm.$error.minlength[0].$invalid || projForm.$error.maxlength[0].$invalid" class="label-form-desc-error">
+                    Description length should be between 50 and 5000 characters.
+                </span>
             </div>
             <div class="text-right">
                 <button class="btn btn-add"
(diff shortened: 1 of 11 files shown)


fnets:
> Running list of things I'm noticing as I'm testing:
> 
> For these two things below, I think you got the limits on the editing part of manage-categories.component.html but not on the adding part. See line 180 in the file.
> 
> * trying to edit a category either in curation view on the project or in the amend pipeline, the error message due to character limit (I'm definitely under minimum) isn't showing up but it won't let me save the category, so the limit is working but it's not showing the error message
> * add category seems to have let me make categories with less-than-minimum description

This should all be addressed now!

SilversunKSauri (approved):
Just tested out the updated category description stuff, works as expected now.  so LGTM now

PR #1047, opened 2023-05-31
Date of Publication issue

van-go:
## Overview: ##
Date of Publication was always showing the newest date, meaning if a version or amend was done, then the Date of Publication was updated. This is an issue because the Date of Publication should always show the actual date if was published. 

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2492](https://jira.tacc.utexas.edu/browse/DES-2492)

## Summary of Changes: ##
For Simulation, FR, Experimental, and Hybrid Sim, I added a dictionary called dateCreated that stores a DOI : dateCreated. In the html, the dictionary is cycled through, and when the entity DOI matches the dictionary DOI, then the dateCreated is displayed. 
For type Other, the Date of Publication is not displayed on the Project side, so no changes were made. 

## Testing Steps: ##
For each project type:
1. Find a project that has versions. 
2. In both  'My Projects' and 'Published' confirm that each entity has it's proper Date of Publication. If it's versioned, these dates should not be all the same. 

For project type other:
1. Confirm that changes are needed.

## UI Photos:
Screen Shot 2023-05-31 at 10 02 20 AM

### DIFF
--- a/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.html
+++ b/designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.html
@@ -99,13 +99,9 @@
                                 (Year of Publication) "{{ mission.value.title }}", in <i>{{ $ctrl.project.value.title }}</i>. DesignSafe-CI.
                                 (DOI)
                             </span>
-                            <span ng-repeat = "list in $ctrl.dateCreated">
-                                <span ng-if="list.key === mission.value.dois.toString()">
-                                    <span ng-if="mission.value.dois.toString() !== ''">
-                                        ({{ list.value | date:'yyyy' }}) "{{ mission.value.title }}", in <i>{{ $ctrl.project.value.title }}</i>. DesignSafe-CI.
-                                        <a href="https://doi.org/{{mission.value.dois.toString()}}">https://doi.org/{{mission.value.dois.toString()}}</a>
-                                    </span>
-                                </span>
+                            <span ng-if="mission.value.dois.toString() !== ''">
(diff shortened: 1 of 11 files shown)


designsafe/static/scripts/data-depot/components/projects/publication-preview/publication-preview-field-recon.component.html
```diff
@@ -189,8 +189,15 @@ <h3 class="text-center">
                                 <div class="entity-meta-data" ng-if="!$ctrl.readOnly && mission.value.dois.toString() === ''">
                                     <span>(Appears when published)</span>
                                 </div>
-                                <div class="entity-meta-data" ng-if="$ctrl.readOnly || mission.value.dois.toString() !== ''">
-                                    <span>{{ $ctrl.dateCreated | date:'MM-dd-yyyy' }}</span>
+                                <div class="entity-meta-data" ng-if="!$ctrl.readOnly && mission.value.dois.toString() !== ''"> 
+                                    <div ng-repeat = "list in $ctrl.dateCreated">
```
jarosenb, line 193: Will there ever be a situation where there are multiple dates associated with one DOI here? If not then `dateList` can probably be an object that looks like `{[doi]: dateCreated}` and you can do an O(1) lookup instead of a loop.

jarosenb (changes requested):

SilversunKSauri (approved):
LGTM, there are a few things that are not taken care of in this PR but we're going to fix the publication-preview on the project side at a later date, so we're making sure we get the published view dates working here, and from my testing it's working as expected.

PR #1078, opened 2023-07-20
Hotfix: use new system monitor

rstijerina:
## Overview: ##

DS was using the old system monitor, causing issues with specific systems. This removes some cases where the form would be disabled, and provides a less scary message if the system is down.

## PR Status: ##

* [X] Ready.

## Testing Steps: ##
1. Go to https://designsafe.dev/rw/workspace/#!/OpenSees::Simulation and select the Stampede2 version of the app
2. Confirm no error message appears
3. Select the Frontera version of the Opensees app
4. Confirm no error message appears


designsafe/static/scripts/workspace/services/systems-service.js
```diff
@@ -3,7 +3,7 @@
  * lookup right now, but it will be nicer to go ahead and code against this service
  * rather than having stubs all over the place.
  */
-export function workspaceSystemsService($q, $http) {
+export default function workspaceSystemsService($q, $http) {
```
jarosenb, line 6:
```suggestion
export function workspaceSystemsService($q, $http) {
```
This either needs to be a non-default export, or anything that imports it needs to be updated to remove the squiggly braces.

jarosenb (changes requested):

jarosenb (approved):
LGTM!

PR #1128, opened 2023-11-15
DES-2564: Project Metadata Pipeline Check

SilversunKSauri:
## Overview: ##
Adding in a check to make sure all required project metadata fields are filled out before a user publishes

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2564](https://jira.tacc.utexas.edu/browse/DES-2564)

## Summary of Changes: ##
New function to check project metadata and added template to display the missing fields

## Testing Steps: ##
1. Create a new project or use an existing project that has not been published.  You will need two, one fr and one type other.
2. Open a python shell, load in both projects and make empty the required fields.
`project.description = ''
project.keywords = ''
project.nh_types = []
#for type other
project.data_type = ''
#for type fr
project.fr_types = []`
3. Save the projects metadatas
4. Go through the publication pipeline and when you hit the "Proofread Project" step, make sure it prompts you to update your project metadata and does not allow you to continue on.
5. Fill out the required fields and make sure that it does allow you to continue on afterwards.

## Notes: ##
I did not test title or PI being empty, because I wasn't sure if clearing those out would break anything else, but neither of those fields has a habit of emptying itself on switching project type.  The other fields are ones that are added to project metadata at that step, and there's a bug where sometimes fields get cleared after switching project type after project creation.  This does not fix that, but it's intended to catch it in the pub pipeline.


van-go (approved):
Tested with type Other and FR. LGTM!

PR #1144, opened 2023-12-16
task/DES-2636: Refactor file move/copy to use non-bulk metadata operations

jarosenb:
## Overview: ##
Refactors bulk metadata operations to use a series of individual create/update calls.
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2636](https://tacc-main.atlassian.net/browse/DES-2636)

## Summary of Changes: ##
- Refactor bulk metadata operations to use synchronous create/update calls on individual metadata objects.
- Make bulk operations asynchronous and allow retries (max 3).
- Update the semantics of `copy_metadata` so that if it needs to be retried, it won't create duplicate metadata records attached to the destination path. (in V3 we can use custom UIDs to make duplication impossible).

## Testing Steps: ##
Perform the following operations on a file with associated metadata (from RApp):
- Copy
- Move
- Rename
- Trash
Ensure that the metadata follows the file across each operation.

### DIFF
--- a/designsafe/apps/api/datafiles/operations/agave_operations.py
+++ b/designsafe/apps/api/datafiles/operations/agave_operations.py
@@ -251,12 +251,12 @@ def move(client, src_system, src_path, dest_system, dest_path):
         filePath=urllib.parse.quote(src_path),
         body=body
     )
-    update_meta(
-        src_system=src_system,
-        src_path=src_path,
-        dest_system=dest_system,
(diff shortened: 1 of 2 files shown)


nathanfranklin (approved):
Excellent! 👍 

We'll shift Hazmapper to use the system/path instead of uuid to find the metadata.

PR #1148, opened 2023-12-21
task/DES-2625: Get publications indexed by Google Scholar

fnets:
## Overview: ##
In order to make it easier for Google Scholar to index DesignSafe publications, I've updated the JSON+LD to look more like an example publication from [EDI](https://portal.edirepository.org/nis/metadataviewer?packageid=knb-lter-fce.1076.4).

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2625](https://tacc-main.atlassian.net/browse/DES-2625)

## Summary of Changes: ##
Changed some structure of the JSON+LD that Google Scholar and Google Dataset use. Made sure that the DOI and Award Numbers are included in a way that is compliant with schema.org.

## Testing Steps: ##
1. Change designsafe.env to use staging (reach out to me on Slack if you need help with this)
2. Go to a random or complicated publication, view the source, and find the JSON in the `<script type="application/ld+json">` tags. 
3. Test it for errors in the [schema.org validator tool](https://validator.schema.org/).
4. Scan for any obvious mistakes, but be aware that the output of the tool doesn't always match the JSON. You will see multiple keys for "hasPart," but really there is one key with multiple value-pairs.


jarosenb (changes requested):
I'm getting an error for PRJ-3853 (chosen at random): `cannot access local variable 'related_ents' where it is not associated with a value`

van-go (commented):
I'm getting this error for all the project I test: `distribution | Dataset is not a known valid target type for the distribution property.`

fnets:
> I'm getting this error for all the project I test: `distribution | Dataset is not a known valid target type for the distribution property.`

I think you may be looking at the wrong branch or a publication that exists in dev. Distribution should no longer be used.

fnets:
> I'm getting an error for PRJ-3853 (chosen at random): `cannot access local variable 'related_ents' where it is not associated with a value`

Good catch! I forgot to reference `self` there. It works on the publication you referenced now.

jarosenb (approved):
This is working now for every publication I've spot-checked!

PR #1170, opened 2024-02-08
Task/des 2679 toolbar trash and rename buttons

van-go:
## Overview: ##
Add Trash Button - user must be logged in (authenticated), more than one file can be moved to the trash. Confirmation pop-ups appear in the bottom left of the screen. 
Add Rename Button - user must be logged in (authenticated), only one file can be renamed at a time. When the rename button is clicked, a Rename modal pops up. The three things I'm validating for are: not-blank, does not allow certain characters (& or @ for example), not the same as the original name. Confirmation pop-ups appear in the bottom left of the screen. 

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2679](https://tacc-main.atlassian.net/browse/DES-2679)
* [DES-2680](https://tacc-main.atlassian.net/browse/DES-2680)

## Summary of Changes: ##
Created a useTrash.ts and useRename.ts
Created a RenameModal.tsx
Updated the DatafilesToolbar.tsx

## Testing Steps: ##
1. Log in using steps for "[React Setup for V3 Development.](https://tacc-main.atlassian.net/wiki/spaces/UP/pages/6659524/Local+Development+Environment+Setup)" The instructions are at the bottom of the wiki page.
2. Trash: Select 1, 2, 3+ files to move to the Trash. Verify they move. Verify the pop-up notification shows for each file. 
3. Rename: Verify you can only select one file at a time to rename. Try breaking the 3 validation rules: blank input, bad characters ($,%,&,etc), and same name as original. When you rename a file, verify that the new name appears and the pop-up show for the file.

## UI Photos:
Screen Shot 2024-02-19 at 2 56 57 PM
Screen Shot 2024-02-19 at 2 57 18 PM


client/modules/datafiles/src/DatafilesModal/RenameModal/RenameModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+    [selectedFiles, mutate, destApi, destSystem, api, system]
+  );
+
+  const handleRenameClick = async () => {
+    const path: string = user?.username ?? '';
+    const newName = form.getFieldValue('newName');
+    updateFileName(newName, path);
+  };
+
+  const handleClose = useCallback(() => {
```
jarosenb, line 57: We can probably remove this `handleClose` callback since unlike previews, there isn't any cached data that we need to get rid of when the modal leaves the screen.

client/modules/_hooks/src/datafiles/useRename.ts
```diff
@@ -0,0 +1,25 @@
+import { useMutation } from '@tanstack/react-query';
+import apiClient from '../apiClient';
+
+type TRenameParam = { api: string; system: string; path: string; name: string };
+
+function renameFn(src: TRenameParam, dest: TRenameParam, doi?: string) {
```
jarosenb, line 6: `src` and `dest` make sense as arguments for move/copy since the system and path could change. For rename, I think there should just be a single param object that has the API, system, path, and new name.

client/modules/datafiles/src/DatafilesToolbar/DatafilesToolbar.tsx
```diff
@@ -34,16 +35,72 @@ export const DatafilesToolbar: React.FC = () => {
...
           selectedFiles.length === 1 && selectedFiles[0].type === 'file',
+        canRename: selectedFiles.length === 1,
         canCopy: user && selectedFiles.length >= 1,
+        canTrash: user && selectedFiles.length >= 1,
       };
     },
     [selectedFiles, user]
   );
 
+  const defaultDestParams = useMemo(
```
jarosenb, line 46: Do we need to set "default" params for trash here? AFAIK it should always respect the api/system/path from the toolbar props.

client/modules/datafiles/src/DatafilesToolbar/DatafilesToolbar.tsx
```diff
@@ -73,6 +130,14 @@ export const DatafilesToolbar: React.FC = () => {
             </ToolbarButton>
           )}
         </DatafilesModal.Copy>
+        <ToolbarButton
```
jarosenb, line 133: Let's extract the trash button into its own component, since it has as much business logic as the rest of the toolbar code combined.

client/modules/datafiles/src/DatafilesModal/RenameModal/RenameModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+            type="textarea"
+            placeholder="Please enter a new name for this file/folder."
+          />
+        </Form.Item>
+      </Form>
+
+      <div style={{ marginTop: '20px', textAlign: 'right' }}>
+        <Button
+          type="primary"
+          onClick={async () => {
```
jarosenb, line 146: Check out the `onFinish` and `onValuesChange` props on the form component: https://ant.design/components/form#api. If you add `htmlType='submit'` to the button, you won't have to define explicit `onClick`/validation methods because the form will handle that for you.

jarosenb (changes requested):

nathanfranklin (changes requested):
Haven't had time to run/test but wanted to leave two quick comments.

client/modules/datafiles/src/DatafilesToolbar/DatafilesToolbar.tsx
```diff
@@ -34,16 +35,36 @@ export const DatafilesToolbar: React.FC = () => {
       return {
```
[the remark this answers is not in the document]
van-go: we can table this for now.

jarosenb (approved):
LGTM, change requests have all been addressed.

PR #1177, opened 2024-03-07
task/DES-2709: Tapis v3 Apps Views

rstijerina:
## Overview: ##

Apps views and updated client call.

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2709](https://tacc-main.atlassian.net/browse/DES-2709)

## UI Images
Screenshot 2024-03-07 at 6 09 09 PM

## Todo
- [x] Tests


designsafe/apps/workspace/api/views.py
```diff
@@ -0,0 +1,293 @@
...
+        return JsonResponse(
+            {
+                "status": 200,
+                "response": data,
+            },
+            encoder=BaseTapisResultSerializer,
+        )
+
+
+@method_decorator(login_required, name="dispatch")
```
jarosenb, line 107: `@login_required` can cause weird redirect behavior, since it's designed for HTML template views. I think we should return a JSON response with a 401 code if the user isn't authenticated.

designsafe/apps/workspace/api/views.py
```diff
@@ -0,0 +1,293 @@
...
+
+        categories.insert(0, {"title": "My Apps", "apps": my_apps})
+
+        # Only return categories that are non-empty
+        categories = list(
+            filter(lambda category: len(category["apps"]) > 0, categories)
+        )
+
+        return JsonResponse(
+            {"categories": categories}, encoder=BaseTapisResultSerializer
```
jarosenb, line 279: What do you think about using Pydantic models to handle representation/serialization of the response here? Advantages would be: 
- provides a formal schema for the response in a more durable way than just documenting it
- automatic case conversion, so we can work in snake_case on the backend and camelCase on the frontend without the linter yelling at us on either side.

designsafe/apps/workspace/api/views.py
```diff
@@ -0,0 +1,293 @@
...
+                    "html",
+                    "icon",
+                    "label",
+                    "license_type",
+                    "popular",
+                    "version",
+                )
+            )
+
+            valid_tapis_apps = self._get_valid_apps(apps_listing, tapis_apps)
```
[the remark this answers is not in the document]
rstijerina: I think we do need to make this call to ensure we're only displaying valid, active apps. Junk and version drift has its way of accumulating in the model admin panel.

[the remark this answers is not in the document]
rstijerina: We do this currently for the Core Portal, it's not a terribly noticeable wait

designsafe/apps/workspace/api/views.py
rstijerina: I haven't used pydantic so hard to comment initially. Case conversion sounds useful, but would we be documenting the Tapis schema? If so that seems a bit heavier than just serializing the response

designsafe/apps/workspace/api/views.py
rstijerina: doesn't `login_required` redirect to login? Is that not desired behavior?

jarosenb: It's desired for an HTML view, but for a JSON endpoint it causes weird behaviors like redirecting to a blob of raw JSON after authentication.

designsafe/apps/workspace/api/views.py
jarosenb: Since we're combining Tapis and database results in these views, I would argue that we're already using a custom response schema, in a sense. Pydantic just makes that explicit. The schemas would be MUCH less involved than the ones we've been using in Projects, because they don't have heterogeneous legacy data to contend with. It would pretty much look like a dataclass with the fields we promise to return.

rstijerina: We'll be returning entire tapipy objects in other calls though, with this same serializer. It sounds like we'd need to spec out the entire object

designsafe/apps/workspace/api/views.py
rstijerina: Hmm I guess I'm not following. Isn't this going to be a template view?

rstijerina: Alright, added a new class `AuthenticatedApiView` which just returns that 401. Lmk what you think
Screenshot 2024-03-07 at 6 22 56 PM

client/modules/workspace/src/lib/workspace.tsx
```diff
@@ -12,11 +13,18 @@ export function Workspace(props: WorkspaceProps) {
       <h1>Welcome to Workspace!</h1>
       {isLoading && <div>Loading app listing...</div>}
       {data && (
-        <ul>
-          {data.map((appMeta) => (
-            <li key={appMeta.uuid}>{appMeta.value.definition.id}</li>
+        <>
+          {data.categories.map((category) => (
+            <div key={uuidv4()}>
```
jarosenb, line 18: Random UUIDs as keys should be a last resort. There should be a way to uniquely identify a category/app based on the API response.

jarosenb (changes requested):

client/modules/workspace/src/lib/workspace.tsx
rstijerina: For sure, this was for speed. I resolved this in the new branch

jarosenb (approved):

PR #1181, opened 2024-03-20
task/DES-2654: v3 Apps Models for main

rstijerina:
## Overview: ##

v3 Apps models for the main branch

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2654](https://tacc-main.atlassian.net/browse/DES-2654)


designsafe/apps/workspace/migrations/0001_initial.py
```diff
@@ -16,8 +16,8 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='AppDescription',
             fields=[
-                ('appId', models.CharField(max_length=120, primary_key=True, serialize=False, unique=True)),
-                ('appDescription', models.TextField(help_text=b'App dropdown description text for apps that have a dropdown.')),
+                ('appid', models.CharField(max_length=120, primary_key=True, serialize=False, unique=True)),
```
rstijerina, line 19: Will these changes be ok? Ideally these would be a separate migration, but I think you mentioned they're already in effect in `main`?

jarosenb: These changes can't be done in a separate migration (I've tried lol). When we moved from MySQL to Postgres, these column names got lowercased in the process. Doing this in a new migration causes the migration planner to construct a SQL statement that renames `appId` to appid`, which fails because `appId` isn't a column anymore. We have to rewrite the history here to tell Django that the column has always been named `appid`.

jarosenb (approved):
Thanks! lgtm

PR #1194, opened 2024-04-01
task/DES-2707: add filemeta app

nathanfranklin:
## Overview: ##

Add filemeta app for storing file metadata. Previously, this was handled by Tapis V2.

The `/api/filemeta/` endpoint will allow users to get and create/update metadata associated with files stored within a DS project’s Tapis system (API limits creation/update of metadata to only files on DS projects).

This endpoint will be used primarily by Rapp user to create metadata for files in that app. Hazmapper and Designsafe will also read these values.

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2704](https://tacc-main.atlassian.net/browse/DES-2704)

## Testing Steps: ##
1. Run unit tests

## Notes: ##
TODO
- [x] unit tests


designsafe/apps/api/filemeta/models.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+from django.contrib.postgres.fields import JSONField
+
+
+class FileMetaModel(models.Model):
+    system = models.CharField(max_length=255)
+    path = models.CharField(max_length=4096)
+    data = models.JSONField()
```
jarosenb, line 8: IMO the `data` field should be renamed to `value` since that's the naming convention in V2 metdata

designsafe/apps/api/filemeta/models.py
```diff
@@ -0,0 +1,17 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
```
jarosenb, line 10:
```suggestion
    created = models.DateTimeField(default=timezone.now)
    last_updated = models.DateTimeField(auto_now=True)
```
Using a "default" for `created` will let us populate it with the dates from the existing metadata, instead of everything having a 2024 creation date.

jarosenb (changes requested):

designsafe/apps/api/filemeta/models.py
nathanfranklin: 👍

jarosenb (approved):

PR #1230, opened 2024-05-07
fix: DES-2765 app variant without label could use app entry/bundle label

wesleyboar:
## Overview: ##

If an app variant does not have a label, it could use the app entry/bundle label.

_This would make "The display name of this app in the Apps Tray. If not defined, uses `notes.label` from app definition." true for someone who misinterprets "Apps Tray" as "Tools & Applications" page grid._

## PR Status: ##

* [x] Ready.
* Work in Progress.
* Hold.

## Related Jira tickets: ##

* [DES-2765](https://tacc-main.atlassian.net/browse/DES-2765)

## Summary of Changes: ##

- **added** `'app'` with `label` to template context for `AppVariants`
- **changed** `app_variant_plugin.html` to use `app.label` as fallback.

## Testing Steps: ##
1. Create/Open a page with an "App Version Selection" instance.
2. Set the plugin instance to an App Entry/Bundle with a variant that has **no** label.
3. Verify the label rendered for the variant is the app's label.

| has App Entry/Bundle label | no Variant label |
| - | - |
| has app label | no variant label |

## UI Photos: ##

| before | after |
| - | - |
| no label shown | use app label |

## Notes: ##

> [!IMPORTANT]
> This is a proposal. Maybe I misunderstand apps.

### DIFF
--- a/designsafe/apps/workspace/cms_plugins.py
+++ b/designsafe/apps/workspace/cms_plugins.py
@@ -100,6 +100,9 @@ class AppVariants(CMSPluginBase):
         context = super().render(context, instance, placeholder)
         app_variants = instance.app.appvariant_set.filter(enabled=True)
         context["listing"] = app_variants
+        context["app"] = {
+            "label": instance.app.label
+        }
 
         return context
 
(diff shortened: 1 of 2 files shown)


wesleyboar:
> I think B would only make sense if the bundle only had one variant.
> — @rstijerina

Designer was saying the same thing. I didn't get it. I will assume multiple variants are assumed to have a different name, because that is how I can understand the statement above.

PR #1247, opened 2024-05-20
task/DES-2755 + DES-2756: Allocations backend + hooks

edmondsgarrett:
## Overview: ##
Adds allocations backend + client hooks for v3 apps redesign
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2755](https://tacc-main.atlassian.net/browse/DES-2755)
* [DES-2756](https://tacc-main.atlassian.net/browse/DES-2756)

## Summary of Changes: ##
- Added allocations API view + TAS helper function for fetching allocations by username
- Added allocations model for elasticsearch + added model to new index in `elasticsearch_settings`
- Added TASAllocation Type for client hooks
- Added `useGetAllocations` client hook for use in apps form

## Testing Steps: ##
1. Go to https://designsafe.dev/api/workspace/allocations and verify the endpoint returns a 200 with a response that looks like the photo below.

## UI Photos:
image

## Notes: ##
- On the form where this hook is used, will need to check allocation shortlist to be used for DS portal against what user has access to.


designsafe/apps/workspace/api/views.py
```diff
@@ -786,3 +831,50 @@ def post(self, request, *args, **kwargs):
...
+        try:
+            if force:
+                logger.info("Forcing TAS allocation retrieval for user:{}".format(username))
+                raise NotFoundError
+            result = {
+                'hosts': {}
+            }
+            result.update(UserAllocations.objects.get(user=user).value)
+            return result
+        except NotFoundError or UserAllocations.DoesNotExist:
```
jarosenb, line 860:
```suggestion
        except (NotFoundError, UserAllocations.DoesNotExist):
```

jarosenb: "except/or" isn't valid Python syntax AFAIK.

jarosenb (changes requested):

designsafe/apps/workspace/api/views.py
```diff
@@ -780,3 +825,50 @@ def post(self, request, *args, **kwargs):
...
+            result = {
+                'hosts': {}
+            }
+            result.update(UserAllocations.objects.get(user=user).value)
+            return result
+        except (NotFoundError, UserAllocations.DoesNotExist):
+            # Fall back to getting allocations from TAS
+            allocations = _get_tas_allocations(username)
+            userAllocs = UserAllocations(user=user, value=allocations)
+            userAllocs.save()
```
rstijerina, line 858: We should also update the allocations cache upon the cached request, otherwise the cache will remain stale and never update. We can kick it off as an async background task so the cached response isn't delayed

designsafe/apps/workspace/models/elasticsearch.py
```diff
@@ -5,20 +5,21 @@
 from django.conf import settings
 from django.db import models
 from elasticsearch_dsl.connections import connections
-from elasticsearch_dsl import (Search, DocType, Date, Nested,
+from elasticsearch_dsl import (Search, Document, Date, Nested,
```
rstijerina, line 8: Are these changes still desired?

rstijerina (approved):

PR #1249, opened 2024-05-20
task/DES-2674 - Versioned Publications aren't being indexed by Google

fnets:
## Overview: ##
Publications that have been versioned aren’t being indexed, due to the way we present metadata and build the URLs. We add v2 , v3, vX to a URL that has been versioned, but the canonical URL that is on the page doesn't include that versioning.  This change now includes only the latest revisions of versioned publications.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2674](https://tacc-main.atlassian.net/browse/DES-2674)

## Summary of Changes: ##
Now includes revision number in sitemap for publications that have been versioned, and only includes the latest revision.

## Testing Steps: ##
1. While on VPN, go to https://designsafe.dev/sitemap.xml
2. Check to make sure that there are some URLs included that end in `v2`, `v3`,`v4`, etc.
3. Make sure that none of these also have other versions included.

### DIFF
--- a/designsafe/sitemaps.py
+++ b/designsafe/sitemaps.py
@@ -182,7 +182,9 @@ class ProjectSitemap(sitemaps.Sitemap):
                     'system' : 'designsafe.storage.published',
                     'revision' : 'v' + str(proj['revision']) if 'revision' in proj else ''
                 }
-                projPath.append('{root}public/{system}/{project}{revision}'.format(**subpath))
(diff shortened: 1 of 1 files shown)


jarosenb:
Closing this-- once we deploy Next, the latest version of each publication will display as the default at `/designsafe.storage.published/PRJ-XXXX`.

PR #1260, opened 2024-05-24
Bugs/DES-2789 Only validate current step on continue

chandra-tacc:
## Overview: ##
Only validate current step on continue. Previous behavior was that it validated all the fields in the form.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2789](https://tacc-main.atlassian.net/browse/DES-2789)

## Summary of Changes: ##
* Remove association of prev and next button onclick to form submit handler. This causes validation of the entire form to trigger. 
    * An alternative approach is to write custom zod resolver, but that complicates things for the final submit button. This seemed to be a simpler solution.
* Use methods#trigger to trigger validation on the fields in the current step.
* z.string() does not check for strings with length 0 in zod (when compared with Yup). Adjust the validation.

## Testing Steps: ##
1. Go to app: https://designsafe.dev/rw/workspace/opensees-express?appVersion=3.6.0
2. Hit continue and it should allow the user to continue to next steps. Previous behavior was to block there and the console log showed error message.

Screenshot 2024-05-24 at 12 25 57 PM

### DIFF
--- a/client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
+++ b/client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
@@ -174,6 +174,11 @@ export const AppsSubmissionForm: React.FC = () => {
   });
   const { handleSubmit, reset, setValue, getValues, watch } = methods;
 
+  // Define type to support calls like method.trigger, which
+  // require literals instead of string or string[]
+  const fieldValues = getValues();
+  type FieldNameUnion = keyof typeof fieldValues;
+
   const getSteps = (): TStep => {
     const formSteps: TStep = {
       configuration: getConfigurationStep(configuration.fields),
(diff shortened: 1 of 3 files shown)


client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
```diff
@@ -319,11 +324,18 @@ export const AppsSubmissionForm: React.FC = () => {
     setSteps(updatedSteps);
   }, [fields]);
 
-  const handleNextStep = useCallback(() => {
-    // setState({ ...state, ...data });
-    const nextPage = steps[current].nextPage;
-    nextPage && setCurrent(nextPage);
-  }, [current]);
+  // Only allow transition to next step, if the current step has
+  // no validation errors.
```
rstijerina, line 328: future note: we should disable 'continue' and 'back' if we know the form is invalid

rstijerina (approved):

client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
chandra-tacc: sounds good, jira to track: [DES-2821](https://tacc-main.atlassian.net/browse/DES-2821)

PR #1265, opened 2024-05-28
task/DES-2726: configure cloud data on login

nathanfranklin:
## Overview: ##

Configure cloud data on login or during `email_confirmation` process.  This also centralizes definition of the to-be-configured systems in one place (common_settings.py).

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2726](https://tacc-main.atlassian.net/browse/DES-2726


designsafe/settings/common_settings.py
```diff
@@ -532,6 +532,12 @@
 AGAVE_USER_STORE_ID = os.environ.get('AGAVE_USER_STORE_ID', 'TACC')
 AGAVE_USE_SANDBOX = os.environ.get('AGAVE_USE_SANDBOX', 'False').lower() == 'true'
 
+AGAVE_SYSTEMS_TO_CONFIGURE = [
```
rstijerina, line 535: New setting, can we use `TAPIS` instead of `AGAVE`?

designsafe/apps/auth/tasks.py
```diff
@@ -47,21 +51,26 @@ def check_or_configure_system_and_user_directory(username, system_id, path):
...
-            tg458981_client.files.setFacl(systemId=system_id,
-                                          path=path,
-                                          operation="ADD",
-                                          recursionMethod="PHYSICAL",
-                                          aclString=f"d:u:{username}:rwX,u:{username}:rwX")
+        if create_path:
+            tg458981_client = get_tg458981_client()
+            try:
+                # User tg account to check if path exists
+                tg458981_client.files.listFiles(systemId=system_id, path=path)
```
rstijerina, line 58: We should also `try / catch` for the `cloud.data` system. I think here we can also be using the user's client to do the listing for all systems

nathanfranklin: The listing of using the user's client occurs on line 39.  If that fails, then we start the process of checking if the directory/path needs to be created/configured (optional) and setting up credentials.

Here (line 58), we're just seeing if the path exists and if doesn't, we create and set acls.

rstijerina: Oops you're right, all good - resolving

rstijerina (approved):

designsafe/apps/auth/views.py
```diff
@@ -64,15 +67,13 @@ def tapis_oauth(request):
...
+    systems_to_configure = get_systems_to_configure(user.username)
     client = user.tapis_oauth.client
 
     for system in systems_to_configure:
         system_id = system["system_id"]
         path = system["path"]
+        create_path = system["create_path"]
         try:
             client.files.listFiles(systemId=system_id, path=path)
             logger.debug(
```
nathanfranklin, line 79: from slack discussions
- [ ] move as much as possible to workers (i.e. 3 systems listings)
- [ ] double check if any other delay exists coming from during login

PR #1271, opened 2024-05-30
Task/DES-2801 [UI] Apps Nav Bar Spacing

sophia-massie:
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2801](https://tacc-main.atlassian.net/browse/DES-2801)

## Summary of Changes: ##
- Removed unused css
- Add to menu theme in main.tsx to match design

## Testing Steps: ##
1. Go to [https://designsafe.dev/rw/workspace/](https://designsafe.dev/rw/workspace/) 
2. Compare spacing to design [here](https://xd.adobe.com/view/bab9a62d-3f2a-45f8-be0f-5680c474d9c1-efea/screen/341f26ad-1618-432e-926a-e8b7055af30b/)

## UI Photos:
Screenshot 2024-05-30 at 9 18 52 AM


client/modules/workspace/src/AppsSideNav/AppsSideNav.tsx
```diff
@@ -127,8 +122,8 @@ export const AppsSideNav: React.FC<{ categories: TAppCategory[] }> = ({
   const defaultKey = `${appId}${appVersion || ''}${currentApp?.bundle_id}`;
 
   return (
-    <div className={styles.appsBrowserSidebar}>
-      <h3>Applications:</h3>
+    <>
+      <h5 style={{ paddingLeft: 50 }}>Applications:</h5>
```
rstijerina, line 126: This works, but since it seems the intent is to center the "Applications" header, are we able to use centering CSS like `justify-items: center`?

client/src/main.tsx
```diff
@@ -32,7 +32,7 @@ const themeConfig: ThemeConfig = {
       colorPrimary: '#1cb500',
     },
     Menu: {
-      itemHeight: 60,
+      itemHeight: 30,
```
rstijerina, line 35: This may also affect the data depot. Can we confirm this only affects the workspace menu?

client/modules/workspace/src/AppsSideNav/AppsSideNav.tsx
```diff
@@ -1,19 +1,14 @@
...
 const AppsNavLink: React.FC<React.PropsWithChildren<{ to: string }>> = ({
   to,
   children,
 }) => {
-  return (
-    <NavLink to={to} className={styles.navLink}>
-      <div>{children}</div>
-    </NavLink>
-  );
+  return <NavLink to={to}>{children}</NavLink>;
```
rstijerina, line 11: With no more `className` prop passed, we can instead get rid of the `AppsNavLink` component and just use the `NavLink` component

client/src/main.tsx
sophia-massie: Yes, I did check around - the only antd Menu in the repo is here in the workspace for apps that I can tell.

client/modules/workspace/src/AppsSideNav/AppsSideNav.tsx
sophia-massie: Done!

client/src/main.tsx
rstijerina, line 35: Using about a 1.25x design pixels to real-life pixels conversion, I think we can bump this to `45` to better meet the proportions in the design

client/modules/workspace/src/AppsSideNav/AppsSideNav.module.css
rstijerina, line 1: Now that this is no longer in use, we can probably get rid of this file and import in AppsSideNav.tsx

client/src/styles.css
```diff
@@ -60,6 +60,11 @@
   right: -20px !important;
 }
 
+.ant-layout-sider-children {
+  border-right: 2px solid var(--global-color-primary--light) !important;
+  overflow: auto;
```
rstijerina, line 65: @wesleyboar Can you remind us how we should be resolving long app labels?
Screenshot 2024-05-30 at 1 39 40 PM

rstijerina: Ah, nevermind. We're going with a `shortLabel` field in an app's `notes`

wesleyboar: Truncate is good for now. I don't what `shortLabel` is, but I trust you.

I'd say **either** the text should wrap **or**, at least, the text is available somewhere else.[^1]

[^1]: Both options are so the user has a way to see enough to distinguish two items that start with similar text.

wesleyboar (approved):
Approved based on screenshots and that the code is less and cleaner. Thanks @rstijerina and @sophia-massie.

rstijerina (approved):
Looks great!

PR #1288, opened 2024-06-07
hotfix: update command to reindex on local development environment

nathanfranklin:
## Overview: ##

Update the management command to reindex ElasticSearch in the dev environment
* refactored so that it doesn't stop if one re-indexing step fails  (the publications reindexing fails currently)
* increased timeout to 300s (was 10s which was causing an issue sometimes)

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2879](https://tacc-main.atlassian.net/browse/DES-2879) - confirming that setup instructions for ReconPortal dev setup works

## Testing Steps: ##
1. get password for ES staging
2. docker exec -it des_django bash
3. `python manage.py es_setup`

### DIFF
--- a/designsafe/apps/data/management/commands/es_setup.py
+++ b/designsafe/apps/data/management/commands/es_setup.py
@@ -29,58 +29,53 @@ class Command(BaseCommand):
         parser.add_argument('--local', help='Remote role to index from, either staging or default', default="dev")
 
     def handle(self, *args, **options):
-        HOSTS = settings.ES_CONNECTIONS[settings.DESIGNSAFE_ENVIRONMENT]['hosts']
         password = getpass.getpass('Enter password for remote ES cluster ')
         username = options.get('username')
         remote = options.get('remote')
         local = options.get('local')
 
-        local_es_client = elasticsearch.Elasticsearch(settings.ES_CONNECTIONS[local]['hosts'])
+        local_es_client = elasticsearch.Elasticsearch(settings.ES_CONNECTIONS[local]['hosts'],
+                                                      timeout=300)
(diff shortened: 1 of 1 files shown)


designsafe/apps/data/management/commands/es_setup.py
```diff
@@ -29,58 +29,53 @@ def add_arguments(self, parser):
...
+                query=query
+            )
+            if not result:
+                failed_indexes.append(index.format(local))
+        logger.info("Finished indexing.")
+        if failed_indexes:
+            logger.error(f"Successfully reindex {len(indexes) - len(failed_indexes)}"
+                         f" of {len(indexes)} indexes\n"
+                         f"failed indexes are: {', '.join(failed_indexes)}\n"
+                         f"(check logs above for errors related to failures)")
```
fnets, line 81: Feel free to push back on this, but since the indexed entities are being printed, it is impossible to scroll up and view the errors. Maybe it would be a good idea to add an option to run without the indexed entities being printed, or to collect the errors in a list like you did with the failed indices?

fnets (approved):
LGTM and works locally, with one minor suggestion

designsafe/apps/data/management/commands/es_setup.py
nathanfranklin: > the indexed entities are being printed, it is impossible to scroll up and view the errors.

the indexed entities aren't being printed; it is just the `designsafe-{}-publications` that fails and the ES library prints out the whole document. but you can pipe to a file to see the error.

fnets: Yeah, makes sense to me

PR #1303, opened 2024-06-17
task/DES-2898: Active job status nav styling

rstijerina:
## Overview: ##
Set background color for Job Status Nav when hovered and when selected.

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-2898](https://tacc-main.atlassian.net/browse/DES-2898)

## Testing Steps: ##
1. Ensure job status nav appears as below when hovered, and selected. Confirm no link cursor when active and hovered while in Job Status.

## UI Photos:

Screenshot 2024-06-17 at 6 30 58 PM

### DIFF
--- a/client/modules/workspace/src/JobStatusNav/JobStatusNav.module.css
+++ b/client/modules/workspace/src/JobStatusNav/JobStatusNav.module.css
@@ -6,3 +6,16 @@
 .badge {
   margin: 0 10px;
 }
+
+.root:hover,
+.root:focus,
+.highlighted-row,
+.highlighted-row:hover,
+.highlighted-row:focus {
+  background-color: #cbdded;
+  text-decoration: none;
+}
+
+.highlighted-row:hover {
+  cursor: unset;
+}
--- a/client/src/styles.css
+++ b/client/src/styles.css
@@ -132,7 +132,7 @@
 }
 
 /* DO scroll app category menu */
-.ant-layout-sider-children > header + div + ul {
+.ant-layout-sider-children > a + div + ul {
   overflow: auto;
   height: 100%;
 }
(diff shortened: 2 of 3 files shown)


sophia-massie (approved):
LGTM!

client/modules/workspace/src/JobStatusNav/JobStatusNav.module.css
```diff
@@ -6,3 +6,16 @@
 .badge {
   margin: 0 10px;
 }
+
+.root:hover,
+.root:focus,
+.highlighted-row,
+.highlighted-row:hover,
+.highlighted-row:focus {
```
wesleyboar, line 14: This code is good. I merely offer new syntax in context.

```suggestion
root:is(:hover, :focus),
.highlighted-row:is(:default, :hover, :focus) {
```

I have used [`:is(…)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) enough to know this works. The [`:default`](https://developer.mozilla.org/en-US/docs/Web/CSS/:default) is new to me, but widely available.

client/src/styles.css
```diff
@@ -132,7 +132,7 @@
 }
 
 /* DO scroll app category menu */
-.ant-layout-sider-children > header + div + ul {
+.ant-layout-sider-children > a + div + ul {
```
wesleyboar, line 135: Are you sure no `.ant-layout-sider-children > header + div + ul` use cases still exist?

wesleyboar (commented):
Thoughts.

client/modules/workspace/src/JobStatusNav/JobStatusNav.module.css
wesleyboar: I removed proposed use of [`:default`](https://developer.mozilla.org/en-US/docs/Web/CSS/:default), which is stated to be only applicable to form elements. The [:is(…)](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) suggestion is valid.

client/src/styles.css
rstijerina: Fairly certain

PR #1322, opened 2024-07-01
Bugs/DES-2872: [AppForm] Handle axios request errors without crashing client

chandra-tacc:
## Overview: ##
Add error boundary over the suspense for app form. This should handle both allocation and getApps suspenses.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2872](https://tacc-main.atlassian.net/browse/DES-2872)

## Summary of Changes: ##
Use ErrorBoundary over suspense, following documentation [here](https://tanstack.com/query/latest/docs/framework/react/guides/suspense/#resetting-error-boundaries)
using react-error-boundary as mentioned above. 

## Testing Steps: ##
1. Go to https://designsafe.dev/rw/workspace/opensees-does-not-exist
2. Once error is seen, hit a valid app in left nav and see that the error disappears. 
3. Any job submission related error are already handled by app form and do not fall into this (tested this by forcing bad data in client code).

## UI Photos:
Screenshot 2024-07-01 at 1 26 53 PM

### DIFF
--- a/client/src/workspace/layouts/AppsViewLayout.tsx
+++ b/client/src/workspace/layouts/AppsViewLayout.tsx
@@ -20,16 +20,19 @@ export const AppsViewLayout: React.FC = () => {
         </div>
       ) : (
         <ErrorBoundary
-          fallbackRender={({ error }) => (
-            <div id="appDetail-wrapper">
-              <Alert
-                message={error?.response?.data?.message ?? error.message}
-                type="error"
-                showIcon
-                style={{ marginTop: 10 }}
-              />
-            </div>
-          )}
+          key={key}
+          fallbackRender={({ error }) =>
+            error && (
+              <div id="appDetail-wrapper">
+                <Alert
+                  message={error?.response?.data?.message ?? error.message}
+                  type="error"
+                  showIcon
+                  style={{ marginTop: 10 }}
+                />
+              </div>
+            )
+          }
         >
           <Suspense
             fallback={
--- a/client/src/workspace/layouts/AppsViewLayout.tsx
+++ b/client/src/workspace/layouts/AppsViewLayout.tsx
@@ -1,6 +1,7 @@
 import React, { Suspense } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
 import { Outlet } from 'react-router-dom';
-import { Layout } from 'antd';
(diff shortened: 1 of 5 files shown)


edmondsgarrett (approved):
LGTM!

client/src/workspace/layouts/AppsViewLayout.tsx
```diff
@@ -18,17 +19,33 @@ export const AppsViewLayout: React.FC = () => {
           {parse(htmlApp.html as string)}
         </div>
       ) : (
-        <Suspense
-          fallback={
-            <Layout>
-              <Spinner />
-            </Layout>
+        <ErrorBoundary
```
rstijerina, line 22: This works 👍 

Curious if you have thoughts on
- Would `errorElement` also work here?
https://reactrouter.com/en/main/route/error-element
- Or error handling within the router? https://tanstack.com/query/latest/docs/framework/react/guides/query-functions#handling-and-throwing-errors

rstijerina (approved):

client/src/workspace/layouts/AppsViewLayout.tsx
chandra-tacc: 1) error-element within router is close to this PR's solution. It provide more protection to overall component instead of just the <Suspense> usage.
I tested it by making changing in workspaceRouter like this

```
{
    id: 'app',
    path: ':appId',
    element: <AppsViewLayout />,
    errorElement: <ErrorBoundary />
},
```
We could put it in the workspaceRouter as a single boundary, which will catch everything.
```
<ErrorBoundary>
    {workspaceRouter}
  </ErrorBoundary>
```

The only change would be extra handling for component specific errors. We are not doing it right now, we are simply showing the error, so doing this might help us with catching all errors.

2) QueryFn error handling - even with error handling here, with suspense usage, we will still need ErrorBoundary. This is because there is no visibility into the error like `isError` 


What do you think? Do you want to move the ErrorBoundary one or two levels up to catch everything?

rstijerina: Either way. This is good enough for me for now. Perhaps we can bring the error boundary up later if need be

PR #1325, opened 2024-07-02
bug/DES-3000: Handle DOI not provided in useCitationMetrics

van-go:
## Overview: ##
In Publications, when you open a publication, the browser makes a few calls to https://designsafeci-next.tacc.utexas.edu/api/publications/data-cite//  (ie no DOI is being sent over), and those calls hang for 30 seconds until they time out. These calls might be tying up the server and causing performance issues.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-3000](https://tacc-main.atlassian.net/browse/DES-3000)

## Summary of Changes: ##
In useCitationMetrics, I added a check that only returns getCitationMetrics if there is a DOI.
 
## Testing Steps: ##
1. With Inspect open, go to Network. 
2. In publications, click on a publication. Check that useCitationMetrics is never called without a DOI

### DIFF
--- a/client/modules/_hooks/src/datafiles/publications/useCitationMetrics.ts
+++ b/client/modules/_hooks/src/datafiles/publications/useCitationMetrics.ts
@@ -55,10 +55,13 @@ export async function getCitationMetrics({
 
 export function useCitationMetrics(doi: string) {
   return useQuery({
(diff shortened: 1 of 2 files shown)


jarosenb (approved):
LGTM

PR #1335, opened 2024-07-08
Task/des 3005 endpoint to preconstructed geojson

sajithalapati16:
## Overview: ##
Create new endpoints to show opentopography data.
## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [DES-3005](https://tacc-main.atlassian.net/browse/DES-3005)

## Summary of Changes: ##
added two new urls -
/opentopodata_center/  & opentopo_coordinates/ which gives the opentopo geojson data constructed using the scripts.


designsafe/apps/rapid/views.py
```diff
@@ -68,6 +68,39 @@ def get_events(request):
     out = [h.to_dict() for h in s[0:total]]
     return JsonResponse(out, safe=False)
 
+def get_opentopodata_center(request):
+    file_path = os.path.join(settings.BASE_DIR, 'designsafe', 'apps', 'rapid', 'opentopography_catalog_of_spatial_boundaries_center_points.geojson')
+    
+    try:
+        with open(file_path, 'r') as file:
```
fnets, line 75: These statements will need to be updated to be Tapipy calls to get the files from Community Data.

fnets (changes requested):
Since the geojsons are going to be put in Community Data, they don't need to be put in the codebase.

designsafe/apps/rapid/views.py
nathanfranklin: can we put that in a separate Jira issue (so would be a separate PR) for getting from Community Data:
```
## Goal

* use tapipy and/or agavepy  that is follow on work this
```

nathanfranklin (changes requested):
This PR looks good. Note that I haven't run it yet locally.

can you add a unit tests (via pytest).  nothing would need to be mocked and you could just check that you get a 200.

My suggestion is that we add getting the file from community data in a second PR.

fnets (approved):
lgtm and ran locally

nathanfranklin (approved):
LGTM 👍

PR #1337, opened 2024-07-08
task/des 3007 Create script to integrate center point geojson with OpenTopo api(OtCatalog) data to construct the complete geojson files.

Beulah-Karrolla:
## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-3007](https://tacc-main.atlassian.net/browse/DES-3007)

## Summary of Changes: ##
* Updated the script to fetch data from OpenTopography OtCatalog API endpoint based on the center point configured from the polygon geometry and adding all the required fields to the geojson data catalog as per the format.


fnets (approved):
lgtm! But we should start thinking about how this will be incorporated into the codebase.

designsafe/apps/rapid/openTopo_Data/scripts/centerpoint_datagenerator.py
```diff
@@ -103,6 +114,36 @@ def save_aggregated_results(output_filepath, all_features):
...
+    # Specify the API endpoint URL
+    api_url="https://portal.opentopography.org/API/otCatalog"
+    # Specify default API query parameters
+    include_federated="false"
+    detail="true"
+    for feature in features:
+        center_point = feature['geometry']['coordinates']
+        minx, maxx, miny, maxy = calculate_bounding_box(center_point)
+        request_url=f'{api_url}?minx={minx}&miny={miny}&maxx={maxx}&maxy={maxy}&detail={detail}&include_federated={include_federated}'
+        response = requests.get(request_url)
```
nathanfranklin, line 130:
```suggestion
        response = requests.get(request_url)
        response.raise_for_status
```

Then you could drop the `if response.status_code == 200:` and the:
```
        else:
            logging.error(f"Failed to fetch data for {feature['properties']['id']}")
```

or if you want to keep the if and else and not use raise_for_status, then the `else` statement should be updated to throw an exception, so that an output file isn't written

```
        else:
            logging.error(f"Failed to fetch data for {feature['properties']['id']}")
            raise Exception("
```

nathanfranklin: ah, is this the situation where there are some geometries in Data_Catalog_Spatial_Boundaries that don't correspond to datasets provided by `API/otCatalog`?

Beulah-Karrolla: Yes, currently the situation is happening for the arcticDEM datasets, but in case of multi-polygon data, there is always a possibility of the centroid not falling in any of the associated polygons. We are yet to figure out handling such exceptions.

designsafe/apps/rapid/openTopo_Data/scripts/centerpoint_datagenerator.py
```diff
@@ -103,6 +114,36 @@ def save_aggregated_results(output_filepath, all_features):
...
+def fetch_otcatalog_api_response(file_path):
+    with open(file_path, 'r') as file1:
+        data = json.load(file1)
+    features = data['features']
+    # Specify the API endpoint URL
+    api_url="https://portal.opentopography.org/API/otCatalog"
+    # Specify default API query parameters
+    include_federated="false"
+    detail="true"
+    for feature in features:
```
nathanfranklin, line 126: I follow how for each Data_Catalog_Spatial_Boundaries dataset, we are making a query to otCatalog.  Have we tried just making a query with otCatalog to get all OT datasets (i.e Latitude from -90 to 90 and longitude from -180 to 180). If that was possible, then a single request is made and would simplify things.

designsafe/apps/rapid/openTopo_Data/scripts/centerpoint_datagenerator.py
nathanfranklin: ah, i see.  hopefully this suggestion would work: https://github.com/DesignSafe-CI/portal/pull/1337/files#r1671274878 

we could also ask OT devs how best to make a single query to get everything.

designsafe/apps/rapid/opentopography_data/scripts/centerpoint_datagenerator.py
```diff
@@ -172,6 +187,28 @@ def save_aggregated_results(output_filepath, all_features):
...
+        dataset = ot_data.get(id)
+        if dataset is not None:
+            feature['properties']['description'] = dataset['description']
+            feature['properties']['doiUrl'] = dataset['url']
+            feature['properties']['host']='OpenTopo' if dataset['identifier']['propertyID']=='opentopoID' else dataset['identifier']['propertyID']
+            feature['properties']['dateCreated'] = dataset['dateCreated']
+            feature['properties']['temporalCoverage'] = dataset['temporalCoverage']
+            feature['properties']['keywords'] = dataset['keywords']
+        else:
+            logging.warning(f"Dataset with ID {feature['properties']['id']} not found in the OpenTopography API response.")
```
nathanfranklin, line 206: Does this occur at the moment?

 If it does occur at the moment do, we need to remove it from the geojson response

If it doesn't occur, do we want to just throw an exception so that script stops and a developer is aware of the issue and can investigate

nathanfranklin (changes requested):
Lgtm, I just have one comment to consider

designsafe/apps/rapid/openTopo_Data/scripts/centerpoint_datagenerator.py
Beulah-Karrolla: Yes, Issue resolved when whole world data is fetched from the API call.

designsafe/apps/rapid/opentopography_data/scripts/centerpoint_datagenerator.py
```diff
@@ -172,6 +198,40 @@ def save_aggregated_results(output_filepath, all_features):
     logging.info(f"Aggregated results saved to: {output_filepath}")
 
 
+def fetch_otcatalog_api_response(output_filepath):
```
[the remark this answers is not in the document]
Beulah-Karrolla: Updated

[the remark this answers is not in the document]
Beulah-Karrolla: Updated script to fetch otcatalog data into the full geometry geojson file

fnets (approved):

taoteg (approved):
LGTM!

Note: You could use the same response data (on line 351) to populate the Django cache for client responses. If you also setup a django management command that runs on a schedule (e.g. daily @ midnight) you could auto-populate both resources, then use the cache for client requests and fall back to the JSON if the cache is stale by checking against the last refresh date (e.g. if refresh was longer than 24 hours ago it is stale).

sajithalapati16:
> LGTM!
> 
> Note: You could use the same response data (on line 351) to populate the Django cache for client responses. If you also setup a django management command that runs on a schedule (e.g. daily @ midnight) you could auto-populate both resources, then use the cache for client requests and fall back to the JSON if the cache is stale by checking against the last refresh date (e.g. if refresh was longer than 24 hours ago it is stale and would indicate something off in the backend).

Yeah, sure that would be great. However, in this script, we are currently using a different approach to populate the data. We are focusing on the GitHub repository uploaded files data as the latest source of truth rather than the API response. We then use the OpenTopography API to create a fully usable dataset for generating the required metadata.

Correct me if I'm wrong, but caching the API response wouldn't make much sense since we are trying to upload these generated data files to the community data and use Tapis to fetch the data from the backend.

That said, we can definitely work on commands to set the scripts to run using Celery or Crontab. I am thinking we could schedule this once we have set up a pipeline to upload the files to the community data. Then, we can set up a job to run at the desired interval to update.

Just to clarify, this is the backup solution we want to fall back to if they don't go with the new OpenTopography API they have exposed with the polygon coordinates to production. We can implement the above-suggested cache solution to eliminate the delay we are facing.

Based on the reply mail from the OpenTopography team to use the GitHub files:
> This does include the https://portal.opentopography.org/apidocs/#/Public/getOtCatalog service, which we are currently updating to include polygon bounds shortly. We have had developers use this in conjunction with the OpenTopography spatial dataset bounds repository (https://github.com/OpenTopography/Data_Catalog_Spatial_Boundaries) as a workaround in the interim."

PR #1349, opened 2024-07-11
Split Datacite events/metrics into separate hooks

jarosenb:
## Overview: ##
Improve page load performance and reduce concurrent uWSGI connections by only fetching datacite event logs when the user opens the relevant modal.


van-go (approved):
Good idea to separate the two! LGTM

PR #1357, opened 2024-07-15
[Job History] Implement expected functionality of "Reuse Inputs"

chandra-tacc:
## Overview: ##
Implements reuse inputs functionality.
Any re-use has to use merge job data with defaults in app form.
There are two approaches of sharing job data with app form.
1. Url param: 
    * pass job uuid in parameter 
    * navigate to app form.
    * App form fetches job data.
    * With this the uuid is visible in the url.
2. Context:  
     * build a context and provider, 
     * wrap the workspace with the provider.
     * In jobs detail and modal, set the context with job data. And navigate to appform.
     * in appform, read job data from context.

I choose 1, because it is visible on what data the appform is working with - url shows the job uuid.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [DES-2843](https://tacc-main.atlassian.net/browse/DES-2843)

## Summary of Changes: ##
* Implement merging of data
* add useJob which is disabled when uuid is not available.
* setup navigation from job to app form.

## Testing Steps: ##
1. Click 'Reuse Inputs' button from Job Listing. Ensure all the related inputs, parameters and configuration is populated.
4. Repeat the above test with 'Reuse Inputs' from Job Detail Modal.
5. Use a bad uuid in url: https://designsafe.dev/rw/workspace/FigureGen-Serial?appVersion=51.0.0&jobUUID=foobar and ensure the app is not broken.
6. Ensure existing app is not broken (no job uuid): https://designsafe.dev/rw/workspace/FigureGen-Serial?appVersion=51.0.0 
7. Ensure reuse button is shown for non-interactive jobs only
8. Ensure reuse button is not shown when job is in progress.

## UI Photos:
Reuse inputs :
Screenshot 2024-07-15 at 10 36 59 AM

Screenshot 2024-07-15 at 10 18 28 AM

Buttons not visible when job is in progress:
Screenshot 2024-07-15 at 10 19 29 AM
Screenshot 2024-07-15 at 10 19 25 AM


client/modules/_hooks/src/workspace/useGetJobs.ts
```diff
@@ -38,6 +38,23 @@ export function useGetJobs(
   return useQuery(getJobsQuery(operation, queryParams));
 }
 
+const getJobQuery = (
+  operation: TJobGetOperations,
+  queryParams: TJobParamsType
+) => ({
+  queryKey: ['workspace', 'getJobs', operation, queryParams],
```
rstijerina, line 45: Should the query key be for `getJob`?
```suggestion
  queryKey: ['workspace', 'getJob', operation, queryParams],
```

client/modules/_hooks/src/workspace/useGetJobs.ts
```diff
@@ -38,6 +38,23 @@ export function useGetJobs(
...
+  queryFn: () => getJobs(operation, queryParams),
+  retry: false,
+  enabled: !!queryParams.uuid,
+});
+
+export function useGetJob(
```
rstijerina, line 51: Is this a fundamentally different request than `useGetJobs`?

client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
```diff
@@ -65,6 +71,9 @@ export const AppsSubmissionForm: React.FC = () => {
   const {
     user: { username },
   } = useAuthenticatedUser() as { user: TUser };
+  const { data: jobData } = useGetJob('select', { uuid: jobUUID }) as {
```
rstijerina, line 74: There's a slight delay in populating the app form while the job data is being fetched. Would it be possible to keep the loading state while this is being requested as well? Might need to be a suspense query?

client/modules/_hooks/src/workspace/useGetJobs.ts
chandra-tacc: the difference is in query, this uses getJobQuery which is disabled if uuid is not available. Whereas, useGetJobs proceeds with an empty uuid and retrieves the full list. We do not want to retrieve full list in this specific usecase. The goal is to only go through with the request if uuid.

client/modules/_hooks/src/workspace/useGetJobs.ts
chandra-tacc: good catch, will fix

client/modules/workspace/src/JobsListing/JobsListing.tsx
```diff
@@ -95,6 +96,15 @@ export const JobsListing: React.FC<Omit<TableProps, 'columns'>> = ({
     markRead: false,
   });
   const { mutate: readNotifications } = useReadNotifications();
+  const navigate = useNavigate();
+  const handleReuseInputs = (job: TTapisJob) => {
```
rstijerina, line 100: Since this is copied here and in `JobsDetailModal`, is there somewhere we can put this as a util function? Or import from the file where it's first declared?

client/modules/_hooks/src/workspace/useGetJobs.ts
chandra-tacc: And I cannot do conditional (!!uuid) retrieval with useGetJobs in the component - react does not like it.

client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
chandra-tacc: good point, will look into switching to suspense.

client/modules/workspace/src/JobsListing/JobsListing.tsx
chandra-tacc: this used useNavigate, so moved it to a shared component instead.

client/modules/workspace/src/AppsSubmissionForm/AppsSubmissionForm.tsx
chandra-tacc: done, moved to suspense

chandra-tacc:
Retested the test scenarios with suspense usage.
only difference is error handling, now with bad uuid fed by the user, an error like this is seen:

Screenshot 2024-07-15 at 1 41 26 PM

rstijerina (approved):

PR #1425, opened 2024-09-03
task/WC-91: Directories in Publication Preview should open in the Curation listing

jarosenb:
## Overview: ##
When users click on directory links inside of the Publication Preview, they should be navigated to that directory in the Curation area.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WC-91](https://tacc-main.atlassian.net/browse/WC-91)

## Testing Steps: ##
1. Go to a project with folders associated to some entity, and click "Prepare to Publish.
2. Check that when you click on a directory name in Publication Preview, you are


van-go (approved):
Pulled down the PR, tested it, works as expected.

PR #1446, opened 2024-09-23
use bulk job to set project ACLs

jarosenb:
## Overview: ##
Addresses an issue where adding multiple users to a project would trigger multiple simultaneous calls to `setfacls`. This could result in race conditions where users' ACLs were not set on the entire project directory.

ACLs are now set using a Tapis job that makes a single `setfacls` call under the hood.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WC-117](https://tacc-main.atlassian.net/browse/WC-117)

## Summary of Changes: ##
- Switch to using a single call to `setfacls` when project members are added/removed.
- Ensure that ``tg458981` has default ACLs on newly created projects.

## Testing Steps: ##
1. Create a new project
2. Check its ACLs on data.tacc to make sure all members **as well as** `tg458981` have default ACLs.


van-go (approved):
LGTM

PR #1477, opened 2024-10-16
Hotfix: Resolve onboarding mkdir race condition

rstijerina:
## Overview: ##

Resolve race condition where the `setfacl` step in onboarding fails because the `mkdir` command has not been run. Instead of resolving each error individually, resolve all each time. No error is thrown if the particular issue has already been resolved and is "re-resolved".

## PR Status: ##

* [X] Ready.


chandra-tacc (approved):
LGTM.
confirmed [tapis api](https://tapis-project.github.io/live-docs/?service=Files#tag/File-Operations/operation/mkdir) is idempotent.

PR #1485, opened 2024-11-13
Quick: Fix args for  RT ticket creation during onboarding

jarosenb:
## Overview: ##
rtpy has the args to `create_ticket` documented incorrectly. Passing `requestors` instead of `requestor` can lead to weird behavior including email confirmations getting lost.


rstijerina (approved):
🙏

PR #1499, opened 2024-12-10
Quick: Prevent null value from being passed as Tapis tracking ID

jarosenb:
## Overview: ##
If tracking ID isn't passed, default to an empty string. Null values produce errors in Tapis.


nathanfranklin (approved):
LGTM 👍 ; not i did not test locally

PR #1521, opened 2025-02-10
DesignSafe: Adjust credential creation to use TMS

chandra-tacc:
## Overview: ##
Changes for TMS Keys migration process.
Adjusted Scenarios:

Onboarding:

- System supporting TMS - credentials are created with createTmsKeys flag set.
- System not supporting TMS - credentials are created with portal generated public and private key.

Job submission

- Create credentials for TMS based systems and try listing again.
- UI will not show push modal for tms key based system. Instead it shows an error status asking to contact support because this is an api error.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WP-832](https://tacc-main.atlassian.net/browse/WP-832)

## Testing Steps: ##
Look for staging tenant in https://stache.utexas.edu/entry/ee36851448345135988a15c0684464ce
and setup design safe and login.
See scenarios in Overview section for testing

1. Remove credentials.
```
from django.contrib.auth import get_user_model
user = get_user_model().objects.get(username='cyemparala')
tapis = user.tapis_oauth.client
tapis.systems.removeUserCredential(systemId='designsafe.storage.default', userName='cyemparala')
```
2. Removing credential to test onboarding
- reset onboarding for system setup. onboarding should work
3. go to an app and submit a job and see that the credentials part works.


designsafe/apps/onboarding/steps/system_access_v3.py
```diff
@@ -132,21 +148,17 @@ def process(self):
...
-            try:
-                create_system_credentials(
-                    self.user.tapis_oauth.client, self.user.username, pub, priv, system
-                )
+                system_definition = self.get_system(system)
+                if system_definition.get("defaultAuthnMethod") != 'TMS_KEYS':
+                    (priv, pub) = createKeyPair()
+                    create_system_credentials_with_keys(
+                        self.user.tapis_oauth.client, self.user.username, pub, priv, system
+                    )
```
rstijerina, line 157: I think we can drop support for the case where TMS_KEYS is false here, since this won't work now that the key service is being retired along with `register_public_key`

designsafe/utils/system.py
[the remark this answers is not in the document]
chandra-tacc: removing.

designsafe/apps/onboarding/steps/system_access_v3.py
chandra-tacc: removed

designsafe/apps/workspace/api/views.py
```diff
@@ -113,18 +117,27 @@ def _get_app(app_id, app_version, user):
...
+        system_id (str): The ID of the Tapis system to test.
+
+    Returns:
+        SystemDef: The system definition if an error occurs.
+    """
+    system_def = tapis.systems.getSystem(systemId=system_id)
     try:
+        # Check if the system uses TMS_KEYS and create credentials if necessary
+        if system_def.get("defaultAuthnMethod") == 'TMS_KEYS':
+            create_system_credentials(tapis, username, system_id, createTmsKeys=True)
```
rstijerina, line 137: I think we want to move this block to the `except` statement below, such that if the `tapis.files.listFiles` fails, then we run this code

chandra-tacc: Yes, adjusted accordingly. In core portal we check twice. list files -> update cred on exception -> list again to confirm.

designsafe/apps/workspace/api/views.py
```diff
@@ -113,18 +117,33 @@ def _get_app(app_id, app_version, user):
...
+    """
+    system_def = tapis.systems.getSystem(systemId=system_id)
     try:
         tapis.files.listFiles(systemId=system_id, path="/")
+        return None
     except (InternalServerError, UnauthorizedError):
-        system_def = tapis.systems.getSystem(systemId=system_id)
+        # Check if the system uses TMS_KEYS and create credentials if necessary
+        if system_def.get("defaultAuthnMethod") == 'TMS_KEYS':
+            create_system_credentials(tapis, username, system_id, createTmsKeys=True)
```
rstijerina, line 140: I think we need to wrap the `try/except` around `create_system_credentials` followed by `listFiles`, since by default the `createSystemCredentials` call will actually throw an exception if the credentials don't work

fnets (commented):
Code LGTM, trying to test locally but having an issue with the onboarding template. Seems unlikely to be related to these changes, but just want to make sure that they work.

client/modules/_hooks/src/datafiles/usePathDisplayName.ts
```diff
@@ -1,8 +1,8 @@
 import { useCallback } from 'react';
 import { useAuthenticatedUser } from '../useAuthenticatedUser';
 
-export const USER_MYDATA_SYSTEM = 'designsafe.storage.default';
-export const USER_WORK_SYSTEM = 'cloud.data';
+export const USER_MYDATA_SYSTEM = 'designsafe.storage.default.tms';
+export const USER_WORK_SYSTEM = 'cloud.data.tms';
```
rstijerina, line 5: NOTE: After testing is completed, and before merging, this will change back:
```suggestion
export const USER_MYDATA_SYSTEM = 'designsafe.storage.default';
export const USER_WORK_SYSTEM = 'cloud.data';
```

rstijerina (changes requested):

rstijerina (approved):

PR #1526, opened 2025-03-05
bug/WP-903: Fix inconsistent behavior when switching publication versions

jarosenb:
## Overview: ##
Fix a bug where switching publication versions would not work correctly when inside of a file listing.

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WP-903](https://tacc-main.atlassian.net/browse/WP-903)

## Summary of Changes: ##
When inside of a listing, use the path (`/PRJ-XXXX`/`/PRJ-XXXXv2`) to determine which version of the publication is being explored.

## Testing Steps: ##
1. Go to a publication with multiple versions (eg PRJ-2933)
2. Change the version and go to a directory within an entity
3. Confirm that the version selector remains on the correct version and the breadcrumbs work correctly.


nathanfranklin (approved):
LGTM 👍 

(left slack messages as curious about what thing)

PR #1530, opened 2025-03-13
task/WP-686: System Status Modal

erikriv16:
## Overview: ##
Implemented a popup modal displaying system statuses and queue information for Frontera, Lonestar6, and Stampede3. The modal is designed similarly to the one available at [TACC Workbench System Status](https://cep.tacc.utexas.edu/workbench/system-status/frontera.tacc.utexas.edu).

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WP-686](https://tacc-main.atlassian.net/browse/WP-686)

## Summary of Changes: ##
Developed a popup modal displaying system status and queue details for Frontera, Lonestar6, and Stampede3. UI similar to TACC Workbench System Status.

## Testing Steps: ##
1.	Navigate to DesignSafe.
2.	Open Tools & Applications.
3.	Click on any application, then select a version.
4.	Hover over  "System Status" in the top-right corner, next to the “View User Guide” link.

## UI Photos:
Screenshot 2025-03-13 at 1 32 43 PM
Screenshot 2025-03-13 at 1 32 53 PM
Screenshot 2025-03-13 at 1 33 03 PM

## Notes: ##
Observing a delay in queue table loading compared to system status updates. Was thinking of optimizing it by having API calls by fetching queue data for all three systems simultaneously, then refreshing the information every minute or something along those lines. Open to any recommendations or suggestions.


client/src/workspace/components/SystemStatusModal/SystemQueueTable.module.css
fnets, line 1: Instead of creating new styling, you might want to copy it from another modal CSS like this one: https://github.com/DesignSafe-CI/portal/blob/main/client/modules/datafiles/src/DatafilesModal/CopyModal/CopyModal.module.css

client/src/workspace/components/SystemStatusModal/SystemStatusModal.module.css
fnets, line 1: Instead of creating new styling, you might want to copy it from another modal CSS like this one: https://github.com/DesignSafe-CI/portal/blob/main/client/modules/datafiles/src/DatafilesModal/CopyModal/CopyModal.module.css

designsafe/apps/api/views.py
```diff
@@ -155,3 +156,41 @@ def post(self, request):
...
+    
+
+
+class SystemMonitorProxyApi(BaseApiView):
+    """"
+    Proxy API for fetching system queue data from Core Portal. 
+    """
+
+    def get(self, request, hostname):
+        CORE_PORTAL_API_BASE = "https://cep.tacc.utexas.edu/api/system-monitor"
```
fnets, line 168: We usually store variables like this in the `common_settings.py` file, so that they can be easily kept track of.  You can see it in the CEP version (we call it `settings.py` there) -- https://github.com/TACC/Core-Portal/blob/c210828226c8c587a76f18e7ae5d8a849adf3a48/server/portal/apps/system_monitor/views.py#L28

designsafe/apps/api/views.py
```diff
@@ -155,3 +156,41 @@ def post(self, request):
...
+            print("Proxy API Error:", str(e)) 
+            return JsonResponse({"error": str(e)}, status=500)
+            
+
+class SystemOverviewProxyApi(BaseApiView):
+    """
+    Proxy API for fetching system monitor overview data from Core Portal. (Load%, Running, Waiting Jobs)
+    """
+    def get(self, request):
+        CORE_PORTAL_API_BASE = "https://cep.tacc.utexas.edu/api/system-monitor"
```
fnets, line 188: We usually store variables like this in the common_settings.py file, so that they can be easily kept track of. You can see it in the CEP version (we call it settings.py there) -- https://github.com/TACC/Core-Portal/blob/c210828226c8c587a76f18e7ae5d8a849adf3a48/server/portal/apps/system_monitor/views.py#L28

fnets (changes requested):

designsafe/apps/api/views.py
```diff
@@ -155,3 +156,41 @@ def post(self, request):
...
+class SystemMonitorProxyApi(BaseApiView):
+    """"
+    Proxy API for fetching system queue data from Core Portal. 
+    """
+
+    def get(self, request, hostname):
+        CORE_PORTAL_API_BASE = "https://cep.tacc.utexas.edu/api/system-monitor"
+
+        url = f"{CORE_PORTAL_API_BASE}/{hostname}"
+        print(f"Proxy API calling: {url}")
```
rstijerina, line 171: Generally we try to avoid print statements. If necessary, you can use `logger.info`, but in this case this won't be needed

designsafe/apps/api/views.py
rstijerina, line 168: We'll want to use https://tap.tacc.utexas.edu/status/ as our source of truth for system status

designsafe/apps/api/views.py
```diff
@@ -155,3 +156,41 @@ def post(self, request):
...
+        CORE_PORTAL_API_BASE = "https://cep.tacc.utexas.edu/api/system-monitor"
+
+        url = f"{CORE_PORTAL_API_BASE}/{hostname}"
+        print(f"Proxy API calling: {url}") 
+
+        
+        try:
+            response = requests.get(url, timeout=10)
+            response.raise_for_status()
+            return JsonResponse(response.json(), safe=False)
```
rstijerina, line 177: Instead of `safe=False`, we can return a dict here of the form `{response: response.json(), status: 200}`

client/src/workspace/components/SystemStatusModal/SystemStatusModal.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { useState } from 'react';
+import { Tooltip, Badge, Spin, Alert } from 'antd';
+import { SystemQueueTable } from './SystemQueueTable';
+import { useSystemOverview } from '../../../hooks/system-status/useSystemOverview';
+import styles from './SystemStatusModal.module.css';
+
+const SystemStatusContent: React.FC = () => {
+  const [activeSystem, setActiveSystem] = useState('frontera'); //default system
```
rstijerina, line 8: Let's set `stampede3` as the default

client/src/hooks/system-status/useSystemOverview.ts
```diff
@@ -0,0 +1,97 @@
...
+export function useSystemOverview() {
+  const [systems, setSystems] = useState<HPCSystem[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    async function fetchSystems() {
+      try {
+        setLoading(true);
+        const response = await fetch('/api/proxy/system-monitor/');
```
rstijerina, line 26: Try creating a hook to handle this request with tanstack query. This will have the added benefit of loading state being built into the hook state.

client/src/workspace/layouts/AppsViewLayout.tsx
```diff
@@ -47,16 +48,20 @@ export const AppsViewLayout: React.FC = () => {
...
-                <a
-                  href={userGuideLink}
-                  target="_blank"
-                  rel="noreferrer"
-                  style={{ marginRight: 10 }}
-                >
-                  View User Guide
-                </a>
-              )}
+              <div style={{ display: 'inline-block' }}>
```
rstijerina, line 51: This probably doesn't need to be wrapped in a div. Also, try moving the `<SystemStatusModal />` to the bottom of this component, as a modal is it's own component.

client/src/workspace/components/SystemStatusModal/SystemStatusModal.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { useState } from 'react';
+import { Tooltip, Badge, Spin, Alert } from 'antd';
```
rstijerina, line 2: Try using the `Modal` component from antd instead of Tooltip

rstijerina (changes requested):

client/src/hooks/system-status/useSystemQueue.js
```diff
@@ -0,0 +1,17 @@
+/*
+  Used in portal/client/src/workspace/SystemQueueTable.tsx 
+  for getting system queue data from core portal
+*/
+export const useSystemQueue = async (hostname) => {
+  const response = await fetch(`/api/proxy/system-monitor/${hostname}/`);
```
rstijerina, line 6: This should also be data fetched from the tap portal. e.g. https://tap.tacc.utexas.edu/status/Stampede3/

client/src/workspace/components/SystemStatusModal/SystemStatusModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <Modal
+      title="System Status"
+      open={isModalVisible}
+      onCancel={onClose}
+      footer={null}
+      width={500}
+    >
+      <div className={styles.modal}>
+        <div className={styles.tabs}>
+          {['frontera', 'lonestar6', 'Stampede3'].map((hostname) => (
```
rstijerina, line 41: Instead of hardcoding these, let's try doing it dynamically with the response data from tap. e.g. from https://tap.tacc.utexas.edu/status/, we can iterate over each key in the response dict, and assume it is of the form of a system object. We can further ensure good data by filtering the response json by checking that `sys?.system_type === 'COMPUTE'`

client/src/hooks/system-status/useSystemOverview.ts
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error(
+      `Request failed: ${response.status} ${response.statusText}`
+    );
+  }
+  const raw = await response.json();
+
+  const data = raw.response;
+
+  return Object.keys(data)
+    .filter((key) => key !== 'Vista')
```
rstijerina, line 30: Was there a requirement to filter out vista? If so, we should create a configurable setting that is an array of strings, which we would use to update the list of systems shown in this modal

designsafe/settings/common_settings.py
```diff
@@ -569,6 +569,12 @@ def guid_filter(record):
 TAPIS_ADMIN_JWT = os.environ.get('TAPIS_ADMIN_JWT')
 TAPIS_TG458981_JWT = os.environ.get('TAPIS_TG458981_JWT')
 
+# Core Portal Configuration
+CORE_PORTAL_API_BASE = os.environ.get("CORE_PORTAL_API_BASE", "https://cep.tacc.utexas.edu/api/system-monitor")
+
+# Tapis API Configuration
```
fnets, line 575:
```suggestion
# TAP API Configuration
```

designsafe/settings/common_settings.py
```diff
@@ -569,6 +569,12 @@ def guid_filter(record):
...
+TAPIS_API_STATUS = os.environ.get("TAP_API_STATUS", "https://tap.tacc.utexas.edu/status/")
```
fnets, line 576:
```suggestion
TAP_API_STATUS = os.environ.get("TAP_API_STATUS", "https://tap.tacc.utexas.edu/status/")
```

designsafe/settings/common_settings.py
```diff
@@ -569,6 +569,12 @@ def guid_filter(record):
 TAPIS_ADMIN_JWT = os.environ.get('TAPIS_ADMIN_JWT')
 TAPIS_TG458981_JWT = os.environ.get('TAPIS_TG458981_JWT')
 
+# Core Portal Configuration
+CORE_PORTAL_API_BASE = os.environ.get("CORE_PORTAL_API_BASE", "https://cep.tacc.utexas.edu/api/system-monitor")
```
fnets, line 573: If you take a look at [how this API is used in the Core Portal](https://github.com/TACC/Core-Portal/blob/a661a0174c921e844fdff84f1c3106b4d4c1c5bc/server/portal/apps/system_monitor/views.py#L28), you'll see that it uses the [same TAP URL ](SYSTEM_MONITOR_URL)that you're setting below. Is there a reason to use this URL, or can you get the same information from the TAP API? 

It's good to use a single source of "truth" for this information, instead of passing it from portal to portal.

designsafe/settings/common_settings.py
fnets: Sorry for the confusion, but TAP is the "TACC Analysis Portal" and Tapis is the name of the API we use for our HPCs.

client/src/workspace/components/SystemStatusModal/SystemQueueTable.module.css
```diff
@@ -0,0 +1,45 @@
+.queueTable {
+  width: 100%;
+  border-collapse: collapse;
+  margin-top: 0;
+
+  /*
+  border-top: 2px solid #ddd;
+  margin-top: 15px;
+  padding: auto;
+  */
```
fnets, line 10:
```suggestion
 
```

fnets (changes requested):
Sal's changes still need to be addressed, plus I added a couple of my own. This is good work!

client/src/hooks/system-status/useSystemOverview.ts
erikriv16: Ok sounds good, I was thinking it wasn't necessary to have because it didn't appear on https://cep.tacc.utexas.edu/workbench/system-status/frontera.tacc.utexas.edu, thats why I did that.

erikriv16: Wait so add Vista as one of the selectable systems on the top of the modal?  Or not?

rstijerina: Yes, leave it as an option. But more specifically, we can show the options dynamically as whatever comes back from the tap response

wesleyboar (commented):
I can't offer review yet. This feature seems to have skipped design consultation. C.J. and I evaluate. This feature might deserve it's own page (like [in Core-Portal](https://cep.tacc.utexas.edu/workbench/system-status)). We'll balance retaining what great work is already here with what final representation could be.

sophia-massie (changes requested):
Adding on to @fnets comment about styling - There is a similar modal in DesignSafe here https://github.com/DesignSafe-CI/portal/blob/c7822c5a3df740dc4b5a2add9090895b8101df6d/client/modules/datafiles/src/projects/modals/MetricsModal.tsx

Take a look at [MetricsModalBody](https://github.com/DesignSafe-CI/portal/blob/c7822c5a3df740dc4b5a2add9090895b8101df6d/client/modules/datafiles/src/projects/modals/MetricsModal.tsx#L49) and [MetricsModal](https://github.com/DesignSafe-CI/portal/blob/c7822c5a3df740dc4b5a2add9090895b8101df6d/client/modules/datafiles/src/projects/modals/MetricsModal.tsx#L378) there. I think using a similar approach using the built in props for different pieces of the modal and avoiding custom classes will keep the same look and feel as the rest of the site. For example, instead of using an h3 for the title, maybe put that in the Modal header prop?

client/src/hooks/system-status/useSystemQueue.js
fnets: Just wanted to echo Sal's comment here

client/src/hooks/system-status/useSystemOverview.ts
fnets: Just wanted to echo Sal's comment here

client/src/hooks/system-status/useSystemQueue.js
fnets: Looks like this still needs to be addressed

client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -28,11 +43,62 @@ export const FormField: React.FC<{
...
+  const [queueData, setQueueData] = useState<QueueItem[]>([]);
+  useEffect(() => {
+    if (selectedSystem) {
+      useSystemQueue(selectedSystem.display_name.toLowerCase())
+        .then((result) => setQueueData(result))
+        .catch(() => setQueueData([]));
+    }
+  }, [selectedSystem]);
+
+  // Attempt to make Stampede3 queue status appear, same async pattern used in useSystemQueuem, currently not working (frontera working fine)...
```
fnets, line 68: Is this code still being used?

client/modules/workspace/src/AppsWizard/AppsFormSchema.ts
```diff
@@ -186,6 +189,27 @@ export const getConfigurationFields = (
     definition.jobAttributes.execSystemId
   ) as TTapisSystem;
 
+  if (definition.jobType === 'BATCH') {
+    const systemOptions = execSystems.map((sys) => ({
+      value: sys.id,
+      label: sys.id,
```
fnets, line 195: What's the difference between value and label here? Is there a more specific name other than value?

client/modules/workspace/src/AppsWizard/AppsFormSchema.ts
```diff
@@ -194,19 +196,24 @@ export const getConfigurationFields = (
...
+    }));
+    const isMulti = systemOptions.length > 1;
+
+    configurationFields.execSystemId = {
       label: 'System',
       name: 'configuration.execSystemId',
       key: 'configuration.execSystemId',
-      required: true,
       type: 'select',
-      options: getAppExecSystems(execSystems).map((q) => ({
```
fnets, line 206: Why did this line get removed?

client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { Button, Form, Input, Select } from 'antd';
+import { Button, Form, Input, Select, Badge } from 'antd';
```
rstijerina, line 2: Badge doesn't appear to be used

client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -85,11 +122,111 @@ export const FormField: React.FC<{
         style={{ textAlign: 'left', marginBottom: description ? 0 : 16 }}
       >
         {type === 'select' ? (
-          <Select
-            {...props}
-            value={getValues(name)}
-            style={{ textAlign: 'left' }}
-          />
+          name === 'configuration.execSystemId' ? (
```
rstijerina, line 125: The aim of this file component is to be as dynamic as possible. Could we try refactoring this out to it's own component? We can also try extending the `Select` element to accept an `after` property, which would be a type `React.FC`, where we can pass the queue status logic

rstijerina (changes requested):
Working great! Some comments

designsafe/apps/api/views.py
```diff
@@ -155,3 +157,39 @@ def post(self, request):
...
+
+    def get(self, request, hostname):
+        url = f"{settings.TAP_API_STATUS}/{hostname}"
+        
+        try:
+            response = requests.get(url, timeout=10)
+            response.raise_for_status()
+            return JsonResponse(response.json(), safe=False)
+        except requests.exceptions.RequestException as e:
+            print("Proxy API Error:", str(e))
```
rstijerina, line 176: Let's use `logger.exception` instead of `print`

designsafe/apps/api/views.py
```diff
@@ -155,3 +157,39 @@ def post(self, request):
...
+    Proxy API for fetching system queue data from TAP. 
+    """
+
+    def get(self, request, hostname):
+        url = f"{settings.TAP_API_STATUS}/{hostname}"
+        
+        try:
+            response = requests.get(url, timeout=10)
+            response.raise_for_status()
+            return JsonResponse(response.json(), safe=False)
```
rstijerina, line 174: Let's not use `safe=False`, but rather return a dict for `JsonResponse` with `{'response': ..., 'status': ...}`

client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -9,6 +9,129 @@ import {
...
+    </div>
+  );
+};
+
+const QueueStatus: React.FC<{
+  value: string;
+}> = ({ value }) => {
+  const { getValues } = useFormContext();
+  const selectedSystemId = getValues('configuration.execSystemId');
+  const [queueData, setQueueData] = useState<TTapisSystemQueue[]>([]);
```
rstijerina, line 87: Once we turn `getSystemQueue` into a hook, we can use the data store from the hook instead of this local state.

client/src/workspace/components/SystemStatusModal/SystemStatusModal.tsx
rstijerina, line 1: We should move this component dir to `client/modules/workspace/src/`

rstijerina (changes requested):
Good progress, getting closer with overall structure

client/modules/workspace/src/SystemStatusNav/SystemStatusNav.tsx
rstijerina, line 1: I'm still seeing the side divider not extending all the way down when apps are collapsed:
Screenshot 2025-06-03 at 5 48 06 PM

client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -85,11 +186,36 @@ export const FormField: React.FC<{
...
+                setValue('configuration.execSystemLogicalQueue', value);
+              }}
+              after={QueueStatus}
+            />
+          ) : (
+            <ExtendedSelect
+              {...props}
+              name={name}
+              value={getValues(name)}
+              style={{ textAlign: 'left', maxWidth: 150, width: '20%' }}
```
rstijerina, line 216: This maxwidth should be removed. We want these fields to extend all the way across
Screenshot 2025-06-03 at 4 47 56 PM

client/modules/workspace/src/components/SystemStatusModal/SystemStatusModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface SystemStatusModalProps {
+  isModalVisible: boolean;
+  onClose: () => void;
+}
+
+export const SystemStatusModal: React.FC<SystemStatusModalProps> = ({
+  isModalVisible,
+  onClose,
+}) => {
+  const [activeSystem, setActiveSystem] = useState('frontera');
```
rstijerina, line 16: The default system should be derived from the selected app. You can use the `useGetAppsSuspense` hook

client/modules/workspace/src/AppsWizard/AppsFormSchema.ts
```diff
@@ -550,6 +556,9 @@ const FormSchema = (
         ? allocations[0]
         : ''
       : '';
+
+    appFields.configuration.defaults['execSystemId'] =
+      definition.jobAttributes.execSystemId;
```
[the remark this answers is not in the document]
rstijerina: Let's remove this here, and make the line on 539, `appFields.configuration.defaults['execSystemId'] = defaultExecSystem.id;`, non-conditional. i.e. remove the if block for that line above
```suggestion
```

client/modules/workspace/src/components/SystemStatusModal/SystemStatusModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+      style={{ top: '230px' }}
+    >
+      <div className={styles.modal}>
+        <div className={styles.tabs}>
+          {systems?.map((sys) => (
+            <button
+              key={sys.display_name}
+              className={`${styles.tabButton} ${
+                activeSystem === sys.display_name ? styles.activeTab : ''
+              }`}
```
rstijerina, line 40: The default system is not shown as selected when the modal first opens
Screenshot 2025-06-03 at 5 46 34 PM

rstijerina (changes requested):
Almost there!

rstijerina (changes requested):

rstijerina (approved):
LGTM, excellent work!

PR #1532, opened 2025-03-24
Task/wc 183 change source for Citations

van-go:
## Overview: ##
Jira ticket: 
We found a more complete source of citations for DOIs in DesignSafe and Ellen wants us to report those numbers. This is two parts and the second can’t be done until the first is finished.
1. Joshua: get API call info to get Clarivate/World of Citation citation count for a DOI. (Then give the ticket to Tracy)
2. Jake or Vanessa: replace current citation count (from DataCite) with Clarivate/World of Science count.
Note: Do not change anything else. All the other usage numbers (downloads, unique requests, etc. should still come from DataCite)

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WC-183](https://tacc-main.atlassian.net/browse/WC-183)

## Summary of Changes: ##

Added API call to Clarivate/World of Citation
Added password to designsafe.env that is used in common_settings.py
Add url to clarivate api

## Testing Steps: ##
1. Find a publication with citation(s)
2. run: curl -X GET "https://api.clarivate.com/apis/wos-starter/v1/documents?db=DRCI&q=DO=< enter DOI here >" \
     -H "X-Apikey: < enter key from Stache here >"
3. In the return, check the WOK value
4. Verify that value is what appears in the "Citation" section in Cite This Data

## UI Photos:
Screen Shot 2025-03-24 at 1 46 00 PM

Screen Shot 2025-03-24 at 1 45 46 PM


designsafe/settings/common_settings.py
```diff
@@ -626,6 +626,10 @@ def guid_filter(record):
 DATACITE_PASS = os.environ.get('DATACITE_PASS')
 DATACITE_SHOULDER = os.environ.get('DATACITE_SHOULDER')
 
+# Clarivate/World of Citation
+WOS_URL = os.environ.get('WOS_URL', 'https://api.clarivate.com/apis/wos-starter/v1/documents')
+WOS_PASS = os.environ.get('WOS_APIKEY')
```
jarosenb, line 631: Can we default this password to an empty string? That way if it ends up missing for whatever reason, it just takes down the Clarivate endpoint and not the entire site.

designsafe/apps/api/publications/operations.py
```diff
@@ -254,3 +256,37 @@ def initilize_publication(publication, status='publishing', revision=None, revis
...
+            source['count'] for source in rspdict['hits'][0]['citations']
+            if source['db'] == 'WOK'
+        ][0]
+
+        return citations
+
+    except requests.exceptions.RequestException as e:
+        return {"error": str(e)}
+    except (KeyError, IndexError) as e:
+        return {"error": "Unexpected response format"}
```
jarosenb, line 292: Can you please format this using the Black formatter? (it's available as a vscode extension)

I'm still trying to figure out how to enforce Python formatting at the CI level without retroactively modifying a million files and destroying git-blame

client/modules/datafiles/src/projects/ProjectCitation/ProjectCitation.tsx
```diff
@@ -7,6 +7,41 @@ import {
 import { MetricsModal } from '../modals/MetricsModal';
 import styles from './ProjectCitation.module.css';
 
+const useClarivateMetrics = (doi: string, shouldFetch = true) => {
```
jarosenb, line 10: This should be refactored to a react-query hook in the `hooks` module so we don't have to maintain our own loading/error states.

jarosenb (changes requested):

designsafe/apps/api/publications/operations.py
van-go: changed to:
`except requests.exceptions.RequestException as e:
        return {
            "error": str(e),
        }
    except (KeyError, IndexError) as e:
        return {
            "error": "Unexpected response format",
        }`

client/modules/datafiles/src/projects/ProjectCitation/ProjectCitation.tsx
van-go: created useClarivateMetrics.ts and added it index.ts

jarosenb (approved):
Thanks for the changes- looks good now

PR #1543, opened 2025-04-22
task/WG-484: Remove two unused endpoints used during prototyping/investigating

nathanfranklin:
## Overview: ##

This PR removes parts of the backend that were used during early prototyping of the OpenTopo integration into the Recon Portal. We now rely on a single endpoint (`/recon-portal/opentopo/`) that uses the official OpenTopo API (which was updated by the OpenTopo team during collaboration with SGX interns). This PR:
- Removes two internal endpoints
- Removes related data fixtures used to generate GeoJSON files (centerpoints and true extents)

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [WG-484](https://tacc-main.atlassian.net/browse/WG-484)


sophia-massie (approved):
LGTM thanks for the cleanup

PR #1558, opened 2025-05-29
task/WG-500: hook for event selection

nathanfranklin:
## Overview: ##

Hook for event selection

## PR Status: ##

* [X] Ready.

## Related Jira tickets: ##

* [WG-500](https://tacc-main.atlassian.net/browse/WG-500)

## Testing Steps: ##
1. Confirm that working as before

## Notes: ##

 - [ ] improve docstring
 - [ ] need context or juts rely on queryparam?
 - [ ] unit test

### DIFF
--- a/client/modules/_hooks/src/reconportal/useSelectedReconPortalEvent.tsx
+++ b/client/modules/_hooks/src/reconportal/useSelectedReconPortalEvent.tsx
@@ -1,0 +1,55 @@
+import React, { createContext, useContext, useState, useEffect } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { type ReconPortalEvents } from '@client/hooks';
+
+type SelectedReconPortalEventContextType = {
+  selectedReconPortalEventIdentfier: string | null;
+  setSelectedReconPortalEventIdentifier: (title: string | null) => void;
+};
+
+const SelectedReconPortalEventContext = createContext<SelectedReconPortalEventContextType | undefined>(undefined);
+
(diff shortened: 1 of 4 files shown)


client/modules/_hooks/src/reconportal/useSelectedReconPortalEvent.tsx
```diff
@@ -0,0 +1,55 @@
...
+        return newParams;
+      });
+      setSelectedReconPortalEventIdentifierState(eventIdentifier);
+    } else {
+      setSearchParams(prev => {
+        const newParams = new URLSearchParams(prev);
+        newParams.delete('eventId');
+        return newParams;
+      });
+      setSelectedReconPortalEventIdentifierState(null);
```
sophia-massie, line 51: When I click on an event, it will switch back to the listing view after of a second or two of showing the event detail in the side panel. It seems like this line is setting the selectedEvent back to null. 

Is there a different way to approach what you expect `setSelectedReconPortalEventIdentifierState(null);` to do at the end of the conditional here?

sophia-massie (changes requested):

client/modules/_hooks/src/reconportal/useSelectedReconPortalEvent.tsx
nathanfranklin: can you recheck. i refactored everything

sophia-massie (approved):
LGTM! Tested and reviewed code 👍

PR #1559, opened 2025-05-30
docs(README): fix designsafe.dev URL link

wesleyboar:
## Overview

Fixed broken link in README.


wesleyboar:
See #1560 instead.

PR #1562, opened 2025-05-30
Task/wc 262 publication symlinks

van-go:
## Overview: ##
Currently, the file selector in the Workspace uses a Tapis file listing on the NHERI/published directory to list the available publications. Once the publication file layout changes are complete, this approach will be problematic because Tapis cannot distinguish whether a symlink points to a file vs. a directory. We can solve this in a straightforward way by creating a database model for known publication symlinks, so that we can look up what a symlink points to based on its path. The deliverables here are:

A database model with a Tapis accessor as the primary key (tapis://<system>/<path>) and the type (dir/file) as a field

Changes to file listings so that if a symlink is detected, we do a database lookup to determine what type  to list it as so that the frontend knows whether to tread it as a file or a dir

## PR Status: ##

* [ ] Ready.
* [X] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WC-262](https://tacc-main.atlassian.net/browse/WC-262)

## Summary of Changes: ##
create a model (PublishedSymlink)
made chagnes to tapis_operations

## Testing Steps: ##
1.


designsafe/apps/api/datafiles/operations/tapis_operations.py
```diff
@@ -43,15 +44,63 @@ def listing(client, system, path, offset=0, limit=100, q=None, *args, **kwargs):
...
+                    type_str = rec.type    # 'file' or 'dir'
+                    format_str = 'folder' if type_str == 'dir' else 'raw'
+                    print(f"Symlink override: {accessor} -> {type_str}")
+                except PublicationSymlink.DoesNotExist:
+                    # Fallback
+                    type_str = 'file'
+                    format_str = 'raw'
+                    print(f"Symlink not found in DB, defaulting to file: {accessor}")
+
+            listing.append({
```
jarosenb, line 74: It's possible we might want to support symlinks in places that aren't published projects. Instead of duplicating all the listing/formatting code inside the conditional block, let's check the `raw_listing` that's already defined in the function (line 90 in this diff) and update its `type` attirbutes if we find a match in the symlink db.

jarosenb (changes requested):

designsafe/apps/api/publications/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Fields:
+        tapis_accessor (str): Unique Tapis path for the symlink (e.g. 'tapis://system/path').
+        type (str): Target type, either 'file' or 'dir'.
+    """
+
+    tapis_accessor = models.CharField(
+        max_length=512, primary_key=True
+    )  # tapis://<system>/<path>
+    type = models.CharField(
```
jarosenb, line 22: We might as well allow a max_length of 255 here in case we ever need to handle different types in the future.

designsafe/apps/api/publications/models.py
```diff
@@ -0,0 +1,27 @@
+"""
+Django model for storing symlink metadata for published Tapis files.
+This table is used to resolve the actual type (file or dir) of symbolic links
+in the published data system, since Tapis cannot always determine the target type.
+"""
+
+from django.db import models
+
+
+class PublicationSymlink(models.Model):
```
jarosenb, line 10: Let's move this model to the `datafiles` app and rename it to `DatafilesSymlinkType`, since it's a generic way to handle symlinks that we might use for other things in the future.

This diff should also have the migration for the model (`python manage.py makemigrations datafiles`)

jarosenb (changes requested):

PR #1600, opened 2025-07-17
task/wg-522 event tag match fix

sophia-massie:
## Overview: ##
Tags on map need to match those in the left panel

    Increase the weight to bold

    Capitalize the event type

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WG-522](https://tacc-main.atlassian.net/browse/WG-522)

## Summary of Changes: ##
Used same styles as used in the nav side bar

## Testing Steps: ##
1. Hover over a Recon Portal dataset and make sure that the style matches the nav bar

## UI Photos:
Screenshot 2025-07-17 at 3 25 00 PM

### DIFF
--- a/client/modules/reconportal/src/LeafletMap/ReconPortalPopUp.tsx
+++ b/client/modules/reconportal/src/LeafletMap/ReconPortalPopUp.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import dayjs from 'dayjs';
 import { ReconPortalEvents } from '@client/hooks';
 import { Tag, Typography, Space } from 'antd';
-import styles from './LeafletMap.module.css';
 import { getReconEventColor } from '../utils';
 
 const { Title, Text } = Typography;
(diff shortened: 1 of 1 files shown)


client/modules/reconportal/src/LeafletMap/ReconPortalPopUp.tsx
```diff
@@ -31,9 +30,13 @@ export const ReconPortalPopup: React.FC<ReconPortalPopuprops> = ({
         </div>
         <Tag
           color={getReconEventColor(dataset)}
-          className={styles.eventDetailTag}
+          style={{
+            fontWeight: 600,
+            fontSize: 14,
+            textTransform: 'capitalize',
+          }}
```
rstijerina, line 37: Can we tell why `styles.eventDetailTag` is not working? Does it need to be more specific? I think a common style would be good in this case, as this same style set is used in `ReconSidePanel`

PR #1617, opened 2025-08-06
Fixed making calls to systems that don't exist

erikriv16:
## Overview: ##
Fixed errors popping up when app with non-TAP system was selected. Issue was caused by system status components trying to fetch queue data for systems that are not valid. 

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WP-1027](https://tacc-main.atlassian.net/jira/software/c/projects/WP/boards/46?assignee=712020%3Aa66df098-eece-4aae-a1c3-6b226867069f&selectedIssue=WP-1027)

## Summary of Changes: ##
- Modified useSystemQueue to prevent API calls when hostname is empty, prevents 404/500 errors when system is invalid to TAP
- Modified FormField - Updated QueueStatus component to use hostname of the selected system rather than displayName, only renders if system exists from systems returned in useSystemOverview

## Testing Steps: ##
1. Navigate to app Hydro-UQ App (Web-Portal)
2. Ensure no errors pop up and queue status component dosen't render

## UI Photos:
Screenshot 2025-08-06 at 4 50 02 PM

### DIFF
--- a/client/modules/workspace/src/AppsWizard/FormField.tsx
+++ b/client/modules/workspace/src/AppsWizard/FormField.tsx
@@ -75,10 +75,14 @@ const QueueStatus: React.FC<{
   const { getValues } = useFormContext();
   const selectedSystemId = getValues('configuration.execSystemId');
   const displayName = getSystemDisplayName(selectedSystemId);
-  const { data: queueData } = useSystemQueue(displayName);
+  const { data: systems } = useSystemOverview();
+  const selectedSystem = systems?.find(
+    (sys) => sys.display_name === displayName
+  );
+  const { data: queueData } = useSystemQueue(selectedSystem?.hostname || '');
   const selectedQueue = queueData?.find((q) => q.name === value);
 
-  if (!value) return null;
+  if (!value || !selectedSystem) return null;
 
   return (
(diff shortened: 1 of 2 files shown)


client/modules/workspace/src/AppsWizard/FormField.tsx
```diff
@@ -75,10 +75,14 @@ const QueueStatus: React.FC<{
   const { getValues } = useFormContext();
   const selectedSystemId = getValues('configuration.execSystemId');
   const displayName = getSystemDisplayName(selectedSystemId);
-  const { data: queueData } = useSystemQueue(displayName);
+  const { data: systems } = useSystemOverview();
+  const selectedSystem = systems?.find(
+    (sys) => sys.display_name === displayName
+  );
```
rstijerina, line 81: Can we key off of `sys.hostname` to match against our actual `selectedSystem` object? We'd have to grab the `currentExecSystem` object from `AppsSubmissionForm`, or find it via `getExecSystemFromId`. This way, we could get the system status for any system with a valid host, rather than depend on our system id being named a certain way.

PR #1622, opened 2025-08-20
task/WC-268: Description Minimum Update

SilversunKSauri:
## Overview: ##
Updating primary descriptions (project and entity) to require 1000 characters

## PR Status: ##

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WC-268](https://tacc-main.atlassian.net/browse/WC-268)

## Testing Steps: ##
1. Check project and entity level descriptions to make sure they require 1000 characters.

### DIFF
--- a/client/modules/datafiles/src/projects/forms/BaseProjectForm.tsx
+++ b/client/modules/datafiles/src/projects/forms/BaseProjectForm.tsx
@@ -384,7 +384,7 @@ export const BaseProjectForm: React.FC<{
       <Form.Item label="Project Description" required>
         What is this project about? How can data in this project be reused? How
         is this project unique? Who is the audience? Description must be between
-        50 and 5000 characters in length.
+        1000 and 5000 characters in length.
         <Form.Item
           name="description"
           rules={[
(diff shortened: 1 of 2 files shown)


SilversunKSauri:
Good catch on the CreateProjectForm.  The Category form is being left as is, since those are not required descriptions according to Maria, even though it is a required field in the form.

client/modules/datafiles/src/projects/forms/CreateProjectForm.tsx
jarosenb, line 1: I think we can actually make the description optional on this initial project creation form. Once they select experiment/simulation/etc they'll run into the limit anyway, so it might make sense to remove the friction point here. IIRC the pre-V3 form had descriptions being optional for a None-type project.

jarosenb (changes requested):

client/modules/datafiles/src/projects/forms/CreateProjectForm.tsx
SilversunKSauri: I'm getting project creation errors from pydantic when I remove the required here.  So for now, I'm just reducing the description mininmum length back to the current 50.

PR #1653, opened 2025-09-22
feat: DSAPP-106 CMS Plugin for User Guide Link

wesleyboar:
## Overview

Add plugin to render User Guide link on CMS app overview page.

## Status

> [!WARNING]
> - [ ] [Prevent false positives.](https://github.com/DesignSafe-CI/portal/pull/1653#discussion_r2770959121)

## Related

* [DSAPP-106](https://tacc-main.atlassian.net/browse/DSAPP-106)

## Changes

- **added** CMS plugin and its model and admin form
- **added** function to derive app from URL path
- **added** function to know whether context is editing Text

## Testing & UI

1. Create/Find an App Listing Entry with an `href` value.
    <sup>E.g. Ground-Motion Portal app. Scroll to "CMS Display Options:".</sup>

    

2. Create/Open page at a path that matches one of an App Listing Entry.
    <sup>E.g. `/use-designsafe/tools-applications/hazard-apps/ground-motion-portal/`.</sup>

    

3. On page, add "User Guide Link" plugin; do **not** choose app.
4. Verify link is rendered that opens page at correct path for User Guide.
    <sup>E.g. `/user-guide/tools/hazard/#scec`.</sup>

	https://github.com/user-attachments/assets/a926492e-6879-40f8-acd9-378abbc58c42

    ✅ Link opens correct page. Video did not capture it.

5. Change page path to one that does not match an app.
6. Verify:
    - link is **not** rendered
    - text "User guide not found." **is** rendered
    - server does **not** crash

    https://github.com/user-attachments/assets/01ce7b8b-eb7a-43c9-acb5-a69ceab7d45d

    ✅ App can not be guessed at wrong path, so no button.

7. Change "User Guide Link" plugin; choose app.
8. Verify button URL is for the app chosen.

	https://github.com/user-attachments/assets/fa3cdd17-410c-4b61-bda0-7fa6e9360a15

    ✅ Link opens correct page. Video did not capture it.

## Notes

The new function `get_instance_from_url` could be used by existing plugins, `RelatedApps` and `AppVariants`, so that their models, `AppVariantsPlugin` and `RelatedAppsPlugin`, need not be used.


designsafe/apps/workspace/cms_plugins.py
```diff
@@ -10,11 +11,30 @@
...
 )
+from designsafe.apps.workspace.forms import AppUserGuideLinkPluginForm
 
 logger = logging.getLogger(__name__)
 
 
+def get_entry_instance(url):
+    """Helper function to get an AppListingEntry instance based on URL."""
+
+    app_listing_entries = AppListingEntry.objects.filter(enabled=True)
```
rstijerina: Are you able to find by the href here, i.e. `AppListingEntry.objects.get(href=url)`?

designsafe/apps/workspace/cms_plugins.py
```diff
@@ -105,3 +127,37 @@ def render(self, context, instance: AppListingEntry, placeholder):
...
+    ):
+        plugin_instance = instance if isinstance(instance, AppUserGuideLinkPlugin) else None
+
+        instance_app = None
+        if isinstance(instance, AppUserGuideLinkPlugin):
+            instance_app = getattr(instance, "app", None)
+        if instance_app is None:
+            instance_app = get_entry_instance(context.get("request").path)
+
+        context = super().render(context, plugin_instance, placeholder)
```
rstijerina: Does this still work if `plugin_instance` is `None`?

designsafe/apps/workspace/cms_plugins.py
wesleyboar: **I think not reliably.**

Using exact match would require:
- data entry be exact (include trailing slash)
- user be on page with URL that does **not** have hash **nor** query params

<details>

The `url` passed is `context.get("request").path`, which can include:

- trailing slashes (e.g. `/use-designsafe/…/adcirc/`)
- query parameters (e.g. `?tab=overview`, `?edit`, `?toolbar_off`)
- hash (e.g. `#specific-element`)

</details>

PR #1708, opened 2026-05-22
bug/WA-529: Fix to Vista version of Jupyter HPC Native

fnets:
## Overview: ##
The Vista version of Jupyter HPC Native had Stampede3 in the title (see screenshot)

## PR Status: ##

* [x] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Jira tickets: ##

* [WA-529](https://tacc-main.atlassian.net/browse/WA-529)

## Summary of Changes: ##
Added version check to findAppById because we are starting to use apps with the same id, but different versions for simplicity and it is causing the wrong apps to be pulled for layout purposes.

## Testing Steps: ##
1. Go to https://designsafe.dev/workspace/jupyter-hpc-native?appVersion=vista
2. You might need to add the app to the [app tray locally](https://designsafe.dev/admin/workspace/applistingentry/)
3. Verify the title is no longer incorrect.

## UI Photos:
Screenshot 2026-05-07 at 10 33 16 AM

### DIFF
--- a/client/modules/workspace/src/utils/apps.ts
+++ b/client/modules/workspace/src/utils/apps.ts
@@ -544,7 +544,6 @@ export const findAppById = (
     for (const app of category.apps) {
       if (app.app_id === appId && app.version === version) {
         return app;
-        }
       }
     }
   }
--- a/client/modules/workspace/src/utils/apps.ts
+++ b/client/modules/workspace/src/utils/apps.ts
@@ -531,17 +531,21 @@ export const useGetAppParams = () => {
  * Find app in app tray categories and get the icon info.
  * @param data TAppCategories or undefined
  * @param appId string - id of an app.
+ * @param version string - version of an app.
  * @returns icon name if available, otherwise null
  */
 export const findAppById = (
   data: TAppCategories | undefined,
-  appId: string
+  appId: string,
+  version: string
 ) => {
   if (!data) return null;
   for (const category of data.categories) {
     for (const app of category.apps) {
       if (app.app_id === appId) {
-        return app;
+        if (app.version === version) {
+          return app;
+        }
       }
     }
   }
--- a/client/modules/workspace/src/utils/apps.ts
+++ b/client/modules/workspace/src/utils/apps.ts
@@ -542,7 +542,7 @@ export const findAppById = (
   if (!data) return null;
   for (const category of data.categories) {
     for (const app of category.apps) {
-      if (app.app_id === appId && app.version === version) {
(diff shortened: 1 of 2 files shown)


client/modules/workspace/src/utils/apps.ts
```diff
@@ -531,17 +531,21 @@ export const useGetAppParams = () => {
...
+  appId: string,
+  version: string
 ) => {
   if (!data) return null;
   for (const category of data.categories) {
     for (const app of category.apps) {
       if (app.app_id === appId) {
-        return app;
+        if (app.version === version) {
+          return app;
```
rstijerina: Could this be `if (app.app_id === appId && app.version === version) {`?

fnets: Good catch! Forgot to clean that up!

client/modules/workspace/src/utils/apps.ts
```diff
@@ -531,16 +531,18 @@ export const useGetAppParams = () => {
...
   data: TAppCategories | undefined,
-  appId: string
+  appId: string,
+  version: string
 ) => {
   if (!data) return null;
   for (const category of data.categories) {
     for (const app of category.apps) {
-      if (app.app_id === appId) {
+      if (app.app_id === appId && app.version === version) {
```
rstijerina: Does this still find the app if the app listing entry does not specify a version?

fnets: I had the same thought, which is why I originally had the statements nested. But, looking at how `portalApp` [is used](https://github.com/DesignSafe-CI/portal/pull/1708/changes/BASE..6160c48deb5bed38103657213336613e8694ac7b#diff-81b5f5a66ce933ab804acb2b9218d4855b93ce8cfda11a5fb2378aecd5f424d4L46), then if the app tray entry doesn't have a specified version the definition of the app should be a suitable replacement.

fnets: But, if we wanted to just rely on the app listing entry as more of a single source of truth, I could remove those fallbacks and use something like 
`if (app.app_id === appId && (!version || app.version === version))` here

rstijerina: Worked great in CEP, maybe bring here too?

rstijerina (approved):

client/modules/workspace/src/utils/apps.ts
fnets: Ha! Thanks for noticing! I made the change, it's working locally for me.

PR #1712, opened 2026-07-09
Re-generate local SSL certs (2026-07-09)

jarosenb:
Update SSL certs for local dev.


SilversunKSauri (approved):
Worked for me locally

PR #1713, opened 2026-07-09
task/WC-461: disable data download button on tombstoned pubs

SilversunKSauri:
## Overview
Disables the data download button if a type other publication is tombstoned, or if **all** entities of a not-type other are tombstoned.

## PR Status

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Links

* [WC-461](https://tacc-main.atlassian.net/browse/WC-461)
* [WC-482](https://tacc-main.atlassian.net/browse/WC-482)

## Summary of Changes
Adds a check to determine if the data download button should be disabled when tombstones are in place

## Testing Steps
1. Go to any publication that is not tombstoned and make sure the data download button is there.
2. Go to any tombstoned publication (type other) and make sure the data download button is not there.
3. Go to PRJ-5755 (you may need to load a more recent db into your local if the mission is not tombstoned) and make sure the data download button is there.

## Notes
Only PRJ-5755 as of this PR has two entities and only one is tombstoned, hence why we're keeping the data download button if not all entities are tombstoned.  We will be restricting access to the tombstoned entity's files via filesystem permissions changes.

### DIFF
--- a/client/src/datafiles/layouts/published/PublishedDetailLayout.tsx
+++ b/client/src/datafiles/layouts/published/PublishedDetailLayout.tsx
@@ -133,7 +133,7 @@ export const PublishedDetailLayout: React.FC = () => {
 
   let entitiesTombstoned = true;
   const entities = data.tree.children;
-  for (let entity of entities) {
+  for (const entity of entities) {
     if (entity.value.tombstone === undefined) {
       entitiesTombstoned = entitiesTombstoned && false;
     } else {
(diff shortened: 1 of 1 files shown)


client/src/datafiles/layouts/published/PublishedDetailLayout.tsx
```diff
@@ -131,6 +131,16 @@ export const PublishedDetailLayout: React.FC = () => {
     (c) => c.value.projectId === projectId
   )?.publicationDate;
 
+  let entitiesTombstoned = true;
```
jarosenb:
```suggestion
  const entitiesTombstoned = entities.some(e => !!e.value.tombstone)
```
I think this will work to find entities with a truthy value for `tombstone`.

jarosenb (changes requested):

client/src/datafiles/layouts/published/PublishedDetailLayout.tsx
SilversunKSauri: updated

jarosenb (approved):

PR #1716, opened 2026-07-15
Task/wp 1196 update django for ds

van-go:
## Overview
Update Django Version
The version being used by the lockfile is Django 4.2.27.
Django 4.2.x remains supported until April 2026.

Recommended target: Django 5.2.12 (latest 5.2 LTS patch).

## PR Status

* [X] Ready.
* [ ] Work in Progress.
* [ ] Hold.

## Related Links

* [WP-1196](https://tacc-main.atlassian.net/browse/WP-1196)

### pyproject.toml
Three dependency constraints changed:

1. Django:
- Before: >=4.2,<5.0.0
- After: exactly 5.2.12

2. django-filer:
- Before: >=2.2,<3.0
- After: >=3.0,<4.0
- Required because filer 2.x explicitly restricts Django to <5.

3. djangocms-text-ckeditor:
- Before: 5.1, which resolved to 5.1.0
- After: exactly 5.1.7
- Fixes its use of Django APIs removed in Django 5.

### poetry.lock
The lockfile was regenerated with Poetry 2.1.3. Principal resolved changes:

1. Django 4.2.27 → 5.2.12
2. django-filer 2.2.8 → 3.5.0
3. djangocms-text-ckeditor 5.1.0 → 5.1.7
4. Added py-svg-hush 0.3.0, required by the newer filer
5. Removed filer dependencies no longer required:
- django-mptt
- django-js-asset
6. Updated package constraints, artifact hashes, and the lockfile content hash

### common_settings.py
Added the compatibility helper import:
`from designsafe.settings.djangocms_forms_compat import apply_storage_compatibility`
Then, after the Django 5 STORAGES configuration:
```
DEFAULT_FILE_STORAGE = STORAGES["default"]["BACKEND"]
apply_storage_compatibility()
```
This lets the archived forms plugin obtain its legacy setting from the authoritative Django 5 storage configuration. It does not introduce a separate storage configuration.

### djangocms_forms_compat.py
This is a new compatibility adapter for `djangocms-forms-maintained`:
```
from django.core.files import storage
from django.utils.module_loading import import_string
```
It conditionally restores the removed `get_storage_class` function:
```
if not hasattr(storage, "get_storage_class"):
    storage.get_storage_class = import_string
```
The conditional makes it safe if the code is ever run with an older Django version where `get_storage_class` still exists.
This adapter is necessary because `djangocms-forms-maintained` still:
- Reads the removed `DEFAULT_FILE_STORAGE` setting
- Imports the removed `django.core.files.storage.get_storage_class`
It should be removed once that archived plugin is replaced with a Django 5-native forms package.

## Testing Steps

1. …


designsafe/settings/common_settings.py
```diff
@@ -291,6 +292,12 @@
         "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
     },
 }
+
+# djangocms-forms-maintained still uses the pre-Django 5 storage API. Keep its
+# configuration sourced from STORAGES until the archived plugin is replaced.
+DEFAULT_FILE_STORAGE = STORAGES["default"]["BACKEND"]
+apply_storage_compatibility()
```
rstijerina: Can we update django-cms as well to not need this compatibility function?

wesleyboar: **I hope so! Please tell me how this goes!** I have long been worried about the form plugin, but [`'django-cms>=3.0'`](https://github.com/avryhof/djangocms-forms/blob/master/setup.py#L51) suggests forms could still work.

> [Why Now Is the Right Time to Migrate from django CMS 3.11](https://www.django-cms.org/en/blog/2026/01/10/why-now-is-the-right-time-to-migrate-from-django-cms-3-11/)
> — Django CMS Blog | Jan. 10, 2026

pyproject.toml
```diff
@@ -88,7 +88,7 @@ package-mode = false
 pytas = {git = "https://bitbucket.org/taccaci/pytas.git", tag = "v1.7.0"}
 attrdict = { git = "https://github.com/DesignSafe-CI/AttrDict", rev = "83b779ee82d5b0e33be695d398162b8f2430ff33" }
 django-recaptcha2 = { git = "https://github.com/DesignSafe-CI/django-recaptcha2", tag = "v1.4.1--compat-django4" }
-djangocms-cascade = { git = "https://github.com/DesignSafe-CI/djangocms-cascade", rev = "9e9f9e3088a0fcfdc1b27ad7e08b68390aa6a570" }
+djangocms-cascade = { git = "https://github.com/van-go/djangocms-cascade.git", rev = "009a7d8ccc5e8b1ff645091fe6780af71a798b31" }
```
wesleyboar: Your `compat/` branches look to be very far behind `master`. Intentional?

Image

designsafe/settings/common_settings.py
van-go: I performed an _exploratory_ upgrade through django CMS 4.1 to 5.0. However, django CMS 4 introduced significant changes to its data model, plugin tree, and publishing architecture. Because migrating our existing content and legacy plugins would require substantially more work and validation, we decided to keep django CMS at version 3.11.10 🤷‍♀️

pyproject.toml
van-go: I ended up not updating the django CMS, so there are no changes to the djangocms-cascade branch.

designsafe/settings/common_settings.py
```diff
@@ -291,6 +291,11 @@
         "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
     },
 }
+
+# djangocms-forms-maintained still reads the legacy setting while using the
+# Django 5 storage registry. Keep it sourced from the canonical configuration.
```
rstijerina, line 296: What does this mean exactly, in particular "Keep it sourced from the canonical configuration."?

jarosenb (approved):
LGTM! CMS/Forms are working locally.

designsafe/settings/common_settings.py
wesleyboar: > Opened #1723 to reword the comment so this is clearer inline.
>
> — Claude (approved by Wesley B)

PR #1726, opened 2026-09-01
BUG: WC-489 Fix listings for dirs with special characters

jarosenb:
## Overview
Add missing URL encoding for listings that contain special characters.


van-go (approved):
LGTM

PR #1735, opened 2026-09-14
Chore: Prevent 500 error on memcached exceptions

jarosenb:
## Overview
Update memcached settings to prevent page load errors when a memcached exception is raised. This has caused transient issues in the past when the EF subsites are redeployed.


SilversunKSauri (approved):
LGTM

